PR #10, opened 2022-07-27
Replace DescriptorTracker with SpkTracker

LLFourn:
This changes the main "tracker" to something that can track any arbitrary list of scripts indexed by an `Ord` type parameter.

This allows you to track more than one descriptor in the same tracker since the tracker is not tied to any descriptor. There are convenience methods for populating a the spk tracker with a descriptor. 

There are also a bunch of other fixes I did while doing this PR included!


LLFourn:
I just pushed a commit adding `KeychainTracker` to provide a safe way of working with a `SpkTracker` where each script pubkey is created using a wildcard descriptor.

PR #14, opened 2022-08-08
A Basic CBF node using nakamoto

rajarshimaitra:
This builds on of #16 ...

Adds the CBF backend using nakamoto..

Adds an example cli app in the same crate bdk-cbf to sync with any given descriptor..


evanlinjin:
Thank you for this PR! This work is being replaced by #153

PR #16, opened 2022-08-12
Core updates as by product of CBF integration.

rajarshimaitra:
This PR pushes the latest changes applied in the CBF integration to bdk_core. 

Note that by default the SparseChain now creates a zero value genesis checkpoint. For this the existing tests were failing.. Updated the test to cater for the default checkpoint.. But Ideally the default should be the genesis blockhash of each network.

Added a new  `StaleReason::CheckpointAlreadyExists` to handle the test  `same_checkpoint_twice_should_be_stale`.. Though I am not sure if we should stale in this case or just quietly move along.. The previous stale reason wasn't being emitted anymore in this case. With previous logic it was happily applying the the duplicate checkpoint..

### DIFF
--- a/bdk_core/tests/test_keychain_tracker.rs
+++ b/bdk_core/tests/test_keychain_tracker.rs
@@ -22,9 +22,11 @@ fn no_checkpoint_and_then_confirm() {
 
     assert_eq!(chain.apply_checkpoint(checkpoint.clone()), ApplyResult::Ok);
 
+    // TODO: Default SparseChain will create one checkpoint with all 0 values.
+    // Convert the default into genesis block of each network.
     assert_eq!(
         chain.iter_checkpoints(..).count(),
-        0,
+        1,
         "adding tx to mempool doesn't create checkpoint"
     );
     assert_eq!(chain.iter_tx().count(), 1);
--- a/bdk_core/tests/test_keychain_tracker.rs
+++ b/bdk_core/tests/test_keychain_tracker.rs
@@ -4,7 +4,7 @@ use bitcoin::OutPoint;
 use checkpoint_gen::{CheckpointGen, ISpec, OSpec, TxSpec};
 
 #[test]
-fn no_checkpoint_and_then_confirm() {
+fn add_single_unconfirmed_tx_and_then_confirm_it() {
     let mut checkpoint_gen = CheckpointGen::new();
     let mut chain = SparseChain::default();
     let mut tracker = KeychainTracker::default();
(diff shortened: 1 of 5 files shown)


rajarshimaitra:
Having trouble figuring out what to do in this test case 
```
---- orphaned_txout_no_longer_appears stdout ----
thread 'orphaned_txout_no_longer_appears' panicked at 'assertion failed: `(left == right)`
  left: `Stale(CheckpointAlreadyExists)`,
 right: `Ok`', bdk_core/tests/test_keychain_tracker.rs:92:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Looking for suggestions..

LLFourn:
Hey @rajarshimaitra I think same checkpoint twice should be idempotent so we should ignore it. Not just same candidate but if candidate has the same hash and height as current latest we should just apply it and return `Ok`.

rajarshimaitra:
Ah yes that makes sense.. I was thinking of it too while writing this extra stale reason.. Seemed too much policing.. I would update accordingly..

rajarshimaitra:
Removed the duplicate checkpoint stale reason.. Now applying the same checkpoint again will return Ok..

bdk_core/tests/test_keychain_tracker.rs
```diff
@@ -22,9 +22,11 @@ fn no_checkpoint_and_then_confirm() {
 
     assert_eq!(chain.apply_checkpoint(checkpoint.clone()), ApplyResult::Ok);
 
+    // TODO: Default SparseChain will create one checkpoint with all 0 values.
+    // Convert the default into genesis block of each network.
```
LLFourn, line 26: I don't think this is a good TODO. What do we gain by doing this?

bdk_core/tests/test_keychain_tracker.rs
```diff
@@ -22,9 +22,11 @@ fn no_checkpoint_and_then_confirm() {
...
     assert_eq!(
         chain.iter_checkpoints(..).count(),
-        0,
+        1,
         "adding tx to mempool doesn't create checkpoint"
```
LLFourn, line 30: this test no longer makes sense. Perhaps just delete for now.

bdk_core/tests/test_sparse_chain.rs
```diff
@@ -235,7 +229,7 @@ fn adding_checkpoint_which_contains_nothing_new_doesnt_create_new_checkpoint() {
         ..Default::default()
     };
     assert_eq!(chain.apply_checkpoint(update.clone()), ApplyResult::Ok);
-    assert_eq!(chain.iter_checkpoints(..).count(), 1);
+    assert_eq!(chain.iter_checkpoints(..).count(), 2);
```
LLFourn, line 232: we need to fix the name of the test since it should create a new checkpoint!

Perhaps we just delete the test for now unless we can think of a better way to test it.

LLFourn (changes requested):
Thanks for this. A few tests no longer make sense and should be deleted I guess.

I will do these changes now.

bdk_core/tests/test_keychain_tracker.rs
LLFourn: I removed the TODO but would like to know in what direction you were thinking?

LLFourn:
I made my own fixes and also added a small feature : `SpkTracker::is_relevant` so you can easily filter out the txs that nakamoto gives you that are not about your wallet.

bdk_core/tests/test_keychain_tracker.rs
rajarshimaitra: I was mostly thinking of if its makes sense to start the sparse chain with 0 value blockhash.. Ideally i felt it should point to the genesis block.. But also as the genesis block will not contain any transactions and we are identifying checkpoints with cumulative transaction hashes it also doesn't make much difference.. So in that light it does make sense to remove the todo..

rajarshimaitra:
> I made my own fixes and also added a small feature : SpkTracker::is_relevant so you can easily filter out the txs that nakamoto gives you that are not about your wallet.

Interesting.. Thanks..  I will get that working in the CBF PR..

PR #17, opened 2022-08-17
WIP: Sparsechain experiment

evanlinjin:
This PR introduces an alternative implementation of `SparseChain` which removes the "checkpoint" concept, and replaces it with the `Delta` object for rollbacks and additions.

The benefits of this are the following:
* When we introduce an older transaction, we only need to rewrite persistent storage from that index (instead of a potential complete rewrite if the tx is included in the first checkpoint).
* TODO


LLFourn:
Outdated.

PR #19, opened 2022-09-04
Coin selection improvements

evanlinjin:
WIP

### DIFF
--- a/bdk_core/src/coin_select.rs
+++ b/bdk_core/src/coin_select.rs
@@ -13,11 +13,9 @@ pub struct CoinSelector<'a> {
 #[derive(Debug, Clone, Copy)]
 pub struct InputCandidate {
     pub value: u64,
-    /// Weight of the "fixed" fields of a `txin` (`prevout` + `nSequence`).
-    /// Typically set as [`TXIN_FIXED_WEIGHT`] for a single Bitcoin input.
-    pub fixed_weight: u32,
-    /// Weight of the "satisfaction" fields of a `txin` (`scriptSig` + `scriptWitness`).
-    pub satisfaction_weight: u32,
+    /// Weight of the `txin`: `prevout` + `nSequence` + `scriptSig` + `scriptWitness`.
+    pub weight: u32,
+    /// Whether this `txin` is spending a segwit output.
     pub has_segwit: bool,
 }
 
@@ -87,10 +85,7 @@ impl CoinSelectorOpt {
 
     /// Fixed weight of the transaction, inclusive of fixed inputs and outputs.
     pub fn fixed_weight(&self) -> u32 {
-        self.fixed_input
-            .map(|fixed_in| fixed_in.fixed_weight)
-            .unwrap_or(0_u32)
-            + self.fixed_additional_weight
+        self.fixed_input.map(|i| i.weight).unwrap_or(0_u32) + self.fixed_additional_weight
     }
 }
 
@@ -122,10 +117,7 @@ impl<'a> CoinSelector<'a> {
             .unwrap_or(0);
 
         self.opts.fixed_weight()
-            + self
-                .selected()
-                .map(|(_, c)| c.satisfaction_weight + c.fixed_weight)
-                .sum::<u32>()
+            + self.selected().map(|(_, c)| c.weight).sum::<u32>()
             + witness_header_extra_weight
     }
 
--- a/bdk_core/src/coin_select.rs
+++ b/bdk_core/src/coin_select.rs
@@ -1,7 +1,7 @@
 use crate::{collections::BTreeSet, Vec};
 use bitcoin::{LockTime, Transaction, TxOut};
 
-const TXIN_BASE_WEIGHT: u32 = (32 + 4 + 4) * 4;
+pub const TXIN_BASE_WEIGHT: u32 = (32 + 4 + 4) * 4;
 
 #[derive(Debug, Clone)]
 pub struct CoinSelector {
@@ -13,6 +13,8 @@ pub struct CoinSelector {
 #[derive(Debug, Clone, Copy)]
 pub struct WeightedCandidate {
     pub value: u64,
+    /// Typically set as [`TXIN_BASE_WEIGHT`] for a single Bitcoin input.
+    pub base_weight: u32,
     pub satisfaction_weight: u32,
     pub is_segwit: bool,
 }
--- a/bdk_core/src/coin_select.rs
+++ b/bdk_core/src/coin_select.rs
@@ -43,6 +43,9 @@ pub struct CoinSelectorOpt {
     pub target_value: u64,
     /// The feerate we should try and achieve in sats per weight unit.
     pub target_feerate: f32,
+    /// The long term feerate if we are to spend an input in the future instead (in sats/wu).
+    /// This is used for calculating waste.
+    pub long_term_feerate: f32,
     /// The minimum absolute fee (in satoshis).
     pub min_absolute_fee: u64,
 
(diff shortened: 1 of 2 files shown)


bdk_core/src/coin_select.rs
```diff
@@ -1,54 +1,63 @@
...
-pub struct WeightedCandidate {
+pub struct InputCandidate {
     pub value: u64,
-    /// Typically set as [`TXIN_BASE_WEIGHT`] for a single Bitcoin input.
-    pub base_weight: u32,
+    /// Weight of the "fixed" fields of a `txin` (`prevout` + `nSequence`).
+    /// Typically set as [`TXIN_FIXED_WEIGHT`] for a single Bitcoin input.
+    pub fixed_weight: u32,
+    /// Weight of the "satisfaction" fields of a `txin` (`scriptSig` + `scriptWitness`).
     pub satisfaction_weight: u32,
```
LLFourn, line 20: can we not just merge `fixed_weight` and `satisfaction_weight` and call it `inclusion_weight`?

evanlinjin: I think we can

evanlinjin: I just called it `weight`, hope that is okay?

bdk_core/src/coin_select.rs
```diff
@@ -1,77 +1,128 @@
...
     pub value: u64,
+    /// Weight of the `txin`: `prevout` + `nSequence` + `scriptSig` + `scriptWitness`.
     pub weight: u32,
-    pub is_segwit: bool,
+    /// Number of inputs contained within this [`InputCandidate`].
+    /// If we are using single UTXOs as candidates, this would be 1.
+    /// If we are working in `OutputGroup`s (as done in Bitcoin Core), this would be > 1.
+    pub input_count: usize,
+    /// Whether this `txin` is spending a segwit output.
+    pub has_segwit: bool,
```
LLFourn, line 23: I think I prefer `is_segwit`.

bdk_core/src/coin_select.rs
```diff
@@ -1,77 +1,128 @@
...
+    pub fixed_input: Option<InputCandidate>,
+    /// The additional fixed weight of the template transaction including: `nVersion`, `nLockTime`
+    /// and fixed `vout`s and the first bytes of `vin_len` and `vout_len`. Weight of `vin` is not
+    /// included.
+    pub fixed_additional_weight: u32,
 }
 
 impl CoinSelectorOpt {
-    pub fn from_weights(base_weight: u32, drain_weight: u32) -> Self {
+    pub fn from_weights(fixed_weight: u32, drain_weight: u32, drain_spend_weight: u32) -> Self {
```
LLFourn, line 66: base seems to be more appropriate since its the starting weight that you add to.

bdk_core/src/coin_select.rs
```diff
@@ -1,77 +1,128 @@
...
+        Self {
+            value: self.value + other.value,
+            weight: self.weight + other.weight,
+            input_count: self.input_count + other.input_count,
+            has_segwit: self.has_segwit || other.has_segwit,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
```
LLFourn, line 40: Can this derive `Copy` too?

bdk_core/src/coin_select.rs
```diff
@@ -1,77 +1,128 @@
...
+pub const TXIN_FIXED_WEIGHT: u32 = (32 + 4 + 4) * 4;
 
 #[derive(Debug, Clone)]
-pub struct CoinSelector {
-    candidates: Vec<WeightedValue>,
+pub struct CoinSelector<'a> {
+    candidates: Vec<InputCandidate>,
     selected: BTreeSet<usize>,
-    opts: CoinSelectorOpt,
+    opts: &'a CoinSelectorOpt,
```
LLFourn, line 10: I think you can just Copy this in each time. I don't think it needs to be a reference.

bdk_core_example/src/main.rs
```diff
@@ -281,21 +282,21 @@ fn main() -> anyhow::Result<()> {
...
-                },
-            );
+            let mut coin_selector = CoinSelector::new(wv_candidates, &coin_selector_opts);
 
             // just select coins in the order provided until we have enough
             let selection = coin_selector.select_until_finished()?;
 
             // get the selected utxos
-            let selected_txos = selection.apply_selection(&candidates).collect::<Vec<_>>();
+            let selected_txos = selection.iter_selected(&candidates).collect::<Vec<_>>();
```
LLFourn, line 299: I prefer `apply_selection` since it is applying it to the original list.

bdk_core/src/coin_select.rs
```diff
@@ -173,13 +233,21 @@ impl CoinSelector {
...
             as u64)
             .max(self.opts.min_absolute_fee);
         let target_fee_without_drain = ((self.opts.target_feerate * base_weight as f32).ceil()
             as u64)
             .max(self.opts.min_absolute_fee);
 
+        let long_term_fee_with_drain = ((self.opts.long_term_feerate * weight_with_drain as f32)
+            .ceil() as u64)
+            .max(self.opts.min_absolute_fee);
+        let long_term_free_without_drain = ((self.opts.long_term_feerate * base_weight as f32)
```
LLFourn, line 247:
```suggestion
        let long_term_fee_without_drain = ((self.opts.long_term_feerate * base_weight as f32)
```

LLFourn: This doesn't make sense to me. Why are we using the `long_term_feerate` against the transaction's weight. The transactions is being broadcast now, not in the "long term". Isn't `long_term_fee_without_drain == target_fee_without_drain`.

bdk_core/src/coin_select.rs
```diff
@@ -196,10 +272,27 @@ impl CoinSelector {
...
+            )
+        };
+
+        let waste = {
+            let base_waste = if use_drain {
+                self.opts.drain_cost()
+            } else {
+                excess
+            };
+            (base_waste + fee) as i64 - long_term_fee as i64
```
LLFourn, line 295: why does the waste go down if the `long_term_fee` goes up??

evanlinjin: I'm sorry, do you mind elaborating?

bdk_core/src/coin_select.rs
evanlinjin: I was getting it confused with "base size" as in [BIP-0141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#block-size). How about `starting_weight`?

bdk_core/src/coin_select.rs
evanlinjin: You are right, only the weight of the inputs should be considered.

bdk_core/src/coin_select.rs
LLFourn: the higher your `long_term_fee` is the more waste there is no? A tx that creates a high long term fee is more wasteful than one that doesn't. If this is correct then this contradicts the calculation here since if `long_term_fee` is higher `waste` is lower. It's possible I am completely misunderstanding the concept here.

bdk_core/src/coin_select.rs
LLFourn: why would you consider the weight of the inputs when their fee is also not being paid in the "long term", they are being paid now. Do you mean outputs?

bdk_core/src/coin_select.rs
evanlinjin: `waste` should be the waste of not spending it now.

bdk_core/src/coin_select.rs
evanlinjin: `waste` is the wasted cost of not spending inputs now, so we want to compare the "cost" of spending inputs now vs later.

LLFourn: ahhh understood thanks

bdk_core/src/coin_select.rs
LLFourn: Ok I think I get it now. I had a different idea in mind. Might be worth a comment to explain the concepts.

evanlinjin:
Replaced by #20

PR #21, opened 2022-09-13
Implement Branch and Bound Coin Selection

evanlinjin:
The implementation of Bnb also resulted in changes to the `CoinSelector` struct.

### DIFF
--- a/bdk_core_example/src/main.rs
+++ b/bdk_core_example/src/main.rs
@@ -7,7 +7,7 @@ use bdk_core::{
         util::sighash::{Prevouts, SighashCache},
         Address, LockTime, Network, Sequence, Transaction, TxIn, TxOut,
     },
-    coin_select::{CoinSelector, CoinSelectorOpt, WeightedValue, TXIN_BASE_WEIGHT},
+    coin_select::{coin_select_bnb, CoinSelector, CoinSelectorOpt, WeightedValue},
     miniscript::{Descriptor, DescriptorPublicKey},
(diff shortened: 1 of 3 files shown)


bdk_core_example/src/main.rs
```diff
@@ -266,7 +266,8 @@ fn main() -> anyhow::Result<()> {
                 .iter()
                 .map(|(plan, utxo)| WeightedValue {
                     value: utxo.value,
-                    weight: plan.expected_weight() as u32,
+                    weight: TXIN_BASE_WEIGHT + plan.expected_weight() as u32,
+                    input_count: 1,
```
LLFourn, line 270: Let's have a helper constructor here to make this very common case easy!

bdk_core_example/src/main.rs
```diff
@@ -287,20 +296,25 @@ fn main() -> anyhow::Result<()> {
                 wv_candidates,
                 CoinSelectorOpt {
                     target_feerate: 0.5,
-                    ..CoinSelectorOpt::fund_outputs(&outputs, &change_output)
+                    min_drain_value: tracker.descriptor(change_keychain).dust_value(),
+                    ..CoinSelectorOpt::fund_outputs(
+                        &outputs,
+                        &change_output,
+                        TXIN_BASE_WEIGHT + change_plan.expected_weight() as u32,
```
LLFourn, line 303: Can we remove `TXIN_BASE_WEIGHT` here and just put it inside?

PR #22, opened 2022-09-21
Allow internal sorting of candidates within `CoinSelector`

evanlinjin:
WIP


evanlinjin:
I don't think this is the right approach.

If we implement sorting internal to `CoinSelector`, it would make it complicated to preselect candidates before sorting. Also we wouldn't be able to use the internal sorting for Bnb (so it limits it's usecase).

I think it will be more elegant if we introduce a new structure: `SelectionPool` (created with `coin_selector.selection_pool(...) -> SelectionPool`) which contains a pool of indexes that the user can sort/filter/select/peek from.

PR #23, opened 2022-09-22
A more flexible Bnb

evanlinjin:
Introduces `BnbIter`, that uses a generic function to determine `BranchStrategy`.


bdk_core/src/coin_select/coin_selector.rs
```diff
@@ -0,0 +1,495 @@
...
+            value,
+            weight,
+            input_count: 1,
+            is_segwit,
+        }
+    }
+
+    /// Effective feerate of this input candidate.
+    /// `actual_value - input_weight * feerate`
+    pub fn effective_value(&self, opts: &CoinSelectorOpt) -> i64 {
```
LLFourn, line 36: just pass in `feerate`

LLFourn (approved):
This is looking like a big improvement. Code is pretty easy to understand and well arranged. The `BranchStrategy` idea is I think the right API.

I will take a crack at improving this next week. I'm happy for it to be merged if/when you're happy with it @evanlinjin.

bdk_core/src/coin_select/bnb.rs
```diff
@@ -0,0 +1,614 @@
+use super::*;
+
+/// Strategy in which we should branch.
+pub enum BranchStrategy<S> {
+    /// We continue exploring subtrees of this node, starting with the inclusion branch.
+    Continue,
```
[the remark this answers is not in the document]
evanlinjin: Thank you for the compliment, I was busting my brains trying to come up with something elegant. That is true, I will do that!

bdk_core/src/coin_select/bnb.rs
```diff
@@ -0,0 +1,651 @@
...
+        BnbLimit::Rounds(rounds) => {
+            bnb.into_iter(&strategy)
+                .take(rounds)
+                .reduce(|b, c| if c.is_some() { c } else { b })
+        }
+        #[cfg(feature = "std")]
+        BnbLimit::Duration(duration) => {
+            let start = std::time::SystemTime::now();
+            bnb.into_iter(&strategy)
+                .take_while(|_| start.elapsed().expect("failed to get system time") <= duration)
```
afilini, line 307: post-merge note: i don't have real numbers but i'm guessing the syscall to get the current time can easily take 10x more time than just running a bnb iteration (due to the context switch), at least on modern cpus.

i think it would make sense to check the time every n iterations, with n probably >= 100. worst case you overshoot a bit, but overall at least you'd spend most of the time actually doing work rather than waiting for a syscall to finish.

there's a catch, though: if you have a very very slow cpu and a lot of utxos you may end up overshooting by a lot. i'm not sure what's the solution here though, i don't think it's worth adding the complexity to measure how long a few iterations take to adjust the `n` parameter. currently the bnb does i think thousands of iterations and nobody ever complained, so `n = 100` i think is more than reasonable imho.

PR #25, opened 2022-10-03
Implement 'drain all'

### LINKED ISSUE #24: Allow optional recipient
`CoinSelectOpt::target_value` should be `Option<u64>`. This way, we can "drain all".

evanlinjin:
Closes #24

### DIFF
--- a/bdk_core/src/coin_select/coin_selector.rs
+++ b/bdk_core/src/coin_select/coin_selector.rs
@@ -317,6 +317,17 @@ impl<'a> CoinSelector<'a> {
                     SelectionConstraint::MinAbsoluteFee,
                     (target_value + self.opts.min_absolute_fee).saturating_sub(selected),
                 ),
+                (
+                    SelectionConstraint::MinDrainValue,
+                    // when we have no target value (hence no recipient txouts), we need to ensure
+                    // the selected amount can satisfy requirements for a drain output (so we at
+                    // least have one txout)
+                    if self.opts.target_value.is_none() {
+                        (fee_with_drain + self.opts.min_drain_value).saturating_sub(selected)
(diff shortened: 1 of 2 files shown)


bdk_core/src/coin_select/coin_selector.rs
```diff
@@ -43,7 +43,8 @@ impl WeightedValue {
 #[derive(Debug, Clone, Copy)]
 pub struct CoinSelectorOpt {
     /// The value we need to select.
-    pub target_value: u64,
+    /// If the value is `None`, we drain everything.
+    pub target_value: Option<u64>,
```
LLFourn, line 47: Isn't this comment wrong? The semantics should just be we don't check this constraint when we `.finish` if it's `None`.

evanlinjin: The way I am thinking about it is this:
* When we don't have a target value, we don't have recipient `txout`s. So then, we MUST have a drain output.
* Hence, for `finish` to succeed with no target value, we need to at least meet the minimum requirements for a drain `txout`.

evanlinjin: Maybe:
```
/// If the value is `None`, we send all non-fee amounts to the drain output.
```

LLFourn: I'm not sure if we want to be that concrete.

```
if the value is `None` then the selection will be complete if it can pay for the drain output and satisfy the other constraints (e.g. minimum fees).
```

Not sure if I'm correct here though. Feel free to just merge this if you are happy with it and I'll take a closer look when I work on this bit.

LLFourn (approved):

PR #35, opened 2022-10-07
Restructure example crate

rajarshimaitra:
Updated the Example Code and many other structural changes..

-  bdk_esplora crate is removed. No dedicated EsploraClient is required, and we can directly work with a esplora_client::BlockingClient. Esplora syncing is now in bdk_core_example/esplora.rs module.

- Similarly added bdk_core_example/rpc.rs module that includes RPC sync. No extra structure. Working purely on bitcoincore-rpc::Client.

Thanks @LLFourn for all the suggestion.. Hopefully I incorporated all of. This is what the current API seems to be fitting together, but can surely be improved more..

The main app contains EsploraSync and RpcSync command to sync via specific interface. Connection config is hardcoded. But it should work.. Tester need to manually set certain config params in the bitcoin.conf file.

This PR is WIP until we finalize the API.

Upcoming work (after API finalization):
 - Add similar modules for electrum and cbf sync
 - Streamline app usage with multi blockchain syncing and broadcasting option (looking for approach suggestions).
 - Persistence (Probably a separate PR, but will be used in the app)
 - An automated testing environment with bitcoind electrsd to perform various functional testing.


bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+fn wallet_name_from_tracker<K: Debug + Clone + Ord>(tracker: &KeychainTracker<K>) -> String {
+    let data = tracker
+        .iter_keychains(..)
+        .fold(Vec::new(), |mut accum, item| {
+            let mut bytes = item.1.to_string().as_bytes().to_vec();
+            accum.append(&mut bytes);
+            return accum;
+        });
+    sha256Hash::hash(&data[..]).to_hex()[0..10].to_string()
+}
```
LLFourn, line 83: I wonder if we just couldn't give it a fixed name in this example.

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+) -> Result<(), RpcError> {
+    let start = std::time::Instant::now();
+
+    // Import spks from Tracker to Client
+    // Can have spks that are already imported
+    let spk_map = &*tracker.script_pubkeys();
+    let _labelled_scripts: Vec<(Option<String>, Script)> = spk_map
+        .iter()
+        .map(|((keychain, _), script)| {
+            let label = Some(format!("{:?}", keychain));
```
LLFourn, line 189: I think this is a nice idea. It should use `{}` rather than debug though. Require that `K` is `Display`.
(but these should be passed in rather than generated here).

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+        if let Ok(addresses) = client.call::<Value>("getaddressesbylabel", &[label]) {
+            let max_index = addresses
+                .as_object()
+                .expect("Address object expected")
+                .iter()
+                .count()
+                - 1;
+            keychain_derive_map.push((keychain, max_index));
+        }
+    }
```
LLFourn, line 218: I think syncing derivation indexes from RPC to client probably belongs in a different function. It would be great if we only passed `sparse_chain` into this function and a list of labels you want to sync against.

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    // Check for reorg. Invalidate checkpoints from reorg height.
+    let mut last_common = None;
+    let best_hash = client.get_best_block_hash()?;
+    for checkpoint in sparse_chain.iter_checkpoints(..).rev() {
+        if checkpoint.hash == client.get_block_hash(checkpoint.height as u64)?
+            && checkpoint.hash != best_hash
+        {
+            last_common = Some(checkpoint);
+        }
+    }
```
LLFourn, line 237: this can more elegantly be done with `find` or `try_find` on the iterator of checkpoints.

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+            debug!("Core wallet is in sync with tracker");
+        }
+    }
+
+    // Check for reorg. Invalidate checkpoints from reorg height.
+    let mut last_common = None;
+    let best_hash = client.get_best_block_hash()?;
+    for checkpoint in sparse_chain.iter_checkpoints(..).rev() {
+        if checkpoint.hash == client.get_block_hash(checkpoint.height as u64)?
+            && checkpoint.hash != best_hash
```
LLFourn, line 233: why is this condition here?

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+        {
+            last_common = Some(checkpoint);
+        }
+    }
+
+    if let Some(checkpoint) = last_common {
+        sparse_chain.invalidate_checkpoint(checkpoint.height);
```
LLFourn, line 240: why are you invalidating the last common one? If they have it in common it is not invalid.

Also invalidate checkpoint is meant to be a private API.

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    let known_confirms = sparse_chain
+        .iter_tx()
+        .filter_map(|(txid, tx_at_block)| {
+            if tx_at_block.confirmation_time.is_some() {
+                Some(txid)
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<Txid>>();
```
LLFourn, line 290: There is no need to do this. You can add redundant data to `sparse_chain`.

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    Ok(client)
+}
+
+/// Sync the Tracker, Sparsechain and Client with each other.
+pub fn rpc_sync<K: Debug + Clone + Ord>(
+    client: &Client,
+    tracker: &mut KeychainTracker<K>,
+    sparse_chain: &mut SparseChain,
+) -> Result<(), RpcError> {
```
LLFourn, line 180: I think api of this function should just be `client: &Client, labels: &[&str], known_tips: impl Iterator<Item = BlockId>`.

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    let known_tips = sparse_chain.iter_checkpoints(..).rev();
+
+    for tip in known_tips {
+        if let Ok(blockhash) = client.get_block_hash(tip.height as u64) {
+            assert!(blockhash == tip.hash);
+            base_tip = Some(tip);
+            break;
+        } else {
+            // TODO: Incase of multi block reorg, handle vector of invalidation
+            invalidate = Some(tip);
```
LLFourn, line 341: Why are we doing invalidation/tip logic here and above. Seems to be duplicating itself.

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+
+    let cp = CheckpointCandidate {
+        transactions,
+        base_tip,
+        invalidate,
+        new_tip,
+    };
+
+    // Sync everything
+    match sparse_chain.apply_checkpoint(cp) {
```
LLFourn, line 353: `sparse_chain` shouldn't be passed in here. The checkpoint should just be returned.

bdk_core/src/sparse_chain.rs
```diff
@@ -547,7 +547,7 @@ impl SparseChain {
     }
 
     // Invalidate all checkpoints after the given height
-    fn invalidate_checkpoint(&mut self, height: u32) {
+    pub fn invalidate_checkpoint(&mut self, height: u32) {
```
LLFourn, line 550: This method is intentionally private. You invalidate things exclusively with a checkpoint.

LLFourn (changes requested):
Thanks @rajarshimaitra. It's looking good. I think the functionality of figuring out the number of derived keys on each label should be a separate function. Other than that I think the APIs need to be cleaned up as I've indicated.

danielabrozzoni:
I still haven't looked at the code, but I was wondering: Bitcoin Core 0.24 includes support for miniscript descriptors, so maybe we should keep importing scripts for Core <0.24, but import descriptors directly if the version is 0.24 or above?

rajarshimaitra:
> I still haven't looked at the code, but I was wondering: Bitcoin Core 0.24 includes support for miniscript descriptors, so maybe we should keep importing scripts for Core <0.24, but import descriptors directly if the version is 0.24 or above?

I thought about doing the descriptor way too, but IIUC bdk_core is intended to handle any random set of spks and track them.. Descriptors being the special case of them.. So in that light decided to keep things spks only for now.. But it would be cool to do the descriptor way directly.. Although not sure what kind of new complexity it will create.. We need to ensure that address tables are in sync between both core and bdk_core..

LLFourn:
If they are using `SpkTracker` then they will have to load `spks` manually like this this example. If the are using `DescriptorTracker` then they can just upload descriptors. We can have examples on both of these.

I think it's better to prioritize showing how to use `bdk_core` (i) by fetching blocks from RPC or by (ii) using RPC as the unspent backend and just using the coin selection. This is the right way for most apps. Storing redundant data locally and "syncing" is a suboptimal developer experience and should only be used for when you really need it. We should have an example for that but out of all of them its the least important one I think.

rajarshimaitra:
> I think it's better to prioritize showing how to use bdk_core (i) by fetching blocks from RPC or by (ii) using RPC as the unspent backend and just using the coin selection.

i) I am interested in chalking out the block extraction example.. That would be something new.. But I think for that to work efficiently we need to store somewhere the `last_sync_height`.. Should this go inside the tracker somewhere?? Or we can do that cool hack that was previously there in bdk rpc of using a dummy address and marking the sync height as its label.. But this does makes the RPC interface stateful in context of a tracker.. Any idea on this?? fetching blocks from start at every sync will surely be very slow..

ii) This approach is very similar to what we have right now?? Or did you mean something else??

evanlinjin:
@rajarshimaitra it should be trivial to store that information externally right? If you find the data structures tedious to work with, let's discuss from there.

rajarshimaitra:
> @rajarshimaitra it should be trivial to store that information externally right? If you find the data structures tedious to work with, let's discuss from there.

Yeah it should be.. I often get confused on what sort of data we want to store in the tracker, and what not.. That also guides the API decisions around these functions.. I think I mostly understand it now and have some plans go forward now that `TxGraph` seems mostly done..

Would talk over it more in the call and take it here forward..

bdk_rpc/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+    Ok(())
+}
+
+/// Initiate a RPC Client for a given configuration and a Tracker.
+pub fn init_client_for_tracker<K: Debug + Clone + Ord>(
+    config: &RpcConfig,
+    tracker: &KeychainTracker<K>,
+) -> Result<Client, RpcError> {
+    let wallet_name = wallet_name_from_tracker(tracker);
```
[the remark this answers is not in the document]
rajarshimaitra: Done..

rajarshimaitra:
Updated the Example Code and many other structural changes.. 

 - `bdk_esplora` crate is removed. No dedicated `EsploraClient` is required, and we can directly work with a `esplora_client::BlockingClient`. Esplora syncing is now in `bdk_core_example/esplora.rs` module.
 
  - Similarly added `bdk_core_example/rpc.rs` module that includes RPC sync. No extra structure. Working purely on `bitcoincore-rpc::Client`. 
  
Thanks @LLFourn for all the suggestion.. Hopefully I incorporated all of. This is what the current API seems to be fitting together, but can surely be improved more.. 
  
The `main` app contains `EsploraSync` and `RpcSync` command to sync via specific interface. Connection config is hardcoded. But it should work.. Tester need to manually set certain config params in the `bitcoin.conf` file. 

This PR is WIP until we finalize the API.

Upcoming work (after API finalization):
 - Add similar modules for `electrum` and `cbf` sync
 - Streamline app usage with multi blockchain syncing and broadcasting option (looking for approach suggestions).
 - Persistence (Probably a separate PR, but will be used in the app)
 - An automated testing environment with `bitcoind` `electrsd` to perform various functional testing.

rajarshimaitra:
Update PR description with the above text..

bdk_core_example/src/esplora.rs
```diff
@@ -0,0 +1,166 @@
...
+            .collect(),
+        last_valid,
+        invalidate,
+        new_tip,
+    };
+
+    Ok((last_active_index, update))
+}
+
+pub fn perform_esplora_sync<K: Debug + Clone + Ord>(
```
LLFourn, line 118: This should be in the same file as `fn main()` (or better yet just in `fn main()`) for the esplora example. The logic in here is the thing you are trying to demonstrate to people so it should be the first thing they see when they open up the example.

bdk_core_example/src/rpc.rs
```diff
@@ -0,0 +1,337 @@
...
+
+    import_client_into_tracker(tracker, client)?;
+    import_tracker_into_client(tracker, client)?;
+    let update = fetch_update_from_client(tx_graph, sparse_chain, client)?;
+    sparse_chain.apply_update(update)?;
+    tracker.sync(&tx_graph);
+
+    debug!("Sync Completed in {} secs", start.elapsed().as_secs());
+    Ok(())
+}
```
LLFourn, line 332: Same as above. This should be in the `fn main()` for the bdk rpc example.

LLFourn (changes requested):
hey @rajarshimaitra thanks for taking the initiative here. Can we also delete `bdk_esplora` repo since I guess you've moved all that here.

I  think we need a different entrypoint for each example (not just different cli commands). I propose:
1) Create `bdk_cli_example` which is a library that processes all the commands
2) Create `bdk_esplora_example`, `bdk_rpc_example` etc which depends on it

The `fn main()` of `bdk_esplora_example` and `bdk_rpc_example` would:

1) call `bdk_cli_example` to extract the descriptors from the environment variables and return the `SparseChain`, `SpkTracker` etc
2) Do a sync *right in the main function* however that is done for the particular chain.
3) Pass off the command execution to `bdk_cli_example` lib.

This allows the two examples to have different dependencies and to really be able to tell how to use that chain. Also `bdk_rpc_example` can have have multiple entry points showing you how to do things.

rajarshimaitra:
Thanks @LLFourn for the suggestion..  Did the following in last push..

 ### New Example Structure
 The existing `bdk_core_example` and `bdk_esplora` crates have been removed.
 In its place three new crates are created
    
 - `bdk_cli_example`: a library crate containing the cli app data structures, and few high level api calls like `get_balacne()`, `create_tx()` etc on with lower level `KeychainTracker`, `SparseChain`, `TxGraph`. This library is a reference of how to "use" the `bdk_core` crate.
    
    
 - `bdk_esplora_example`: a binary crate that uses `bdk_cli_example` and creates a minimal viable cli wallet app connected to an Eslora backend. This library demonstrate how to connect the internals of `bdk_core` to Esplora API for blockchain data.
    
    
 - `bdk_rpc_example`: Same binary ali app as `bdk_esplora_example`, connects with an Bitcoin Core RPC instance for blockchain data.

Let me know if this fits along with your idea.. I think it worked pretty well in separating the dependencies in their own silos.. 

I thought of removing `rpc.rs` module. But was becoming too much code in `main.rs`.. So kept it there..

rajarshimaitra:
Added electrum sync example.. Very similar to Esplora.. Reorg detection logic similar to rpc (this can be applied to esplora too)..

rajarshimaitra:
Upcoming works:
 - CBF example with nakamoto
 - Basic tests with `bitcoind`, `elctrsd`.
 - RPC/Electrum syncing via block subscriptions.  
 - Add persistence when #51 lands.. 
 
 I think its okay for this PR to have conflict with master.. I will rebase it finally when bdk_core is stabilizes.. For now its seems enough to write out for examples..

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,277 @@
...
+            new_tip,
+        },
+    ))
+}
+
+pub fn perform_electrum_sync<K: Debug + Clone + Ord>(
+    client: &Client,
+    tracker: &mut KeychainTracker<K>,
+    chain: &mut SparseChain,
+    graph: &mut TxGraph,
```
LLFourn, line 157: In the examples we don't want to have these things be mutable while doing IO. The point we want to make is that you can get stuff from the wallet and then drop mutable references, fetch data without any mutable references open and then later get a mutable reference to apply the result. I know for CLI we don't even need to do this because there's only one thread but let's see if we can structure stuff so it's like this. I know the example I wrote originally doesn't do this but see if you can get this message across when designing it.

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,277 @@
...
+    };
+
+    // Construct a Electrum Client
+    use electrum_client::Config;
+    let config = Config::default();
+    let client = Client::from_config("ssl://electrum.blockstream.info:60002", config)?;
+    // Sync at boot
+    perform_electrum_sync(&client, &mut tracker, &mut sparse_chain, &mut tx_graph)?;
+
+    match args.command {
```
LLFourn, line 238: Can't this whole thing just go in the example cli library like "process_cmd" and pass in the references above?

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,277 @@
...
+        .map(|descriptor| Descriptor::<DescriptorPublicKey>::parse_descriptor(&secp, &descriptor))
+        .transpose()?;
+
+    let change_keychain = if let Some((internal_descriptor, internal_keymap)) = internal {
+        keymap.extend(internal_keymap);
+        tracker.add_keychain(Keychain::Internal, internal_descriptor);
+        Keychain::Internal
+    } else {
+        Keychain::External
+    };
```
LLFourn, line 229: can't all of the above be done in the example cli library?

LLFourn (commented):
Hey @rajarshimaitra, The structure looks much better. I agree that it's going to be hard to merge this while we are still making changes. I think there'll be one or two more significant changes to the API and then we should focus on merging and refining all these examples.

I also did a new esplora example in [#bdk](https://github.com/bitcoindevkit/bdk/pull/793). It turns out we might be able to use the same "example" for bdk and bdk core. Or at least I think from the blockchain side they will look very very similar eventually. Best to do the examples as you have been here for now though since the BDK side will probably have more friction.

bdk_electrum_example/src/main.rs
rajarshimaitra: Okay.. Ya I think I see what you are saying.. So we will have some "applier" to insert data inside this structs?

bdk_electrum_example/src/main.rs
rajarshimaitra: Hmm ya.. Could be.. Makes sense to be there actually.. Don't remember why I put it here..

bdk_electrum_example/src/main.rs
rajarshimaitra: Totally..

bdk_electrum_example/src/main.rs
rajarshimaitra: I think I will understand it more once I understand the current core structures.. Seems a lot has changed..

evanlinjin:
To carry this PR forward, I propose the following:
1. Wait until #63 is complete.
2. Make a "move only" PR and merge it ASAP.
3. Continue the rest of the work in a separate PR.

Would this make sense? Thanks!

rajarshimaitra:
By a brief look of https://github.com/LLFourn/bdk_core_staging/pull/63 it seems there's a lot of conflict with this PR. 

As this PR is restructuring the crates if we have #63 first it will end up cherry picking those changes back into this restructuring before merge..

Its always messy when big PRs are open for a long time. But if we are agreeing on the restructuring then it makes sense to first get the restructuring done, and then add extra stuffs into it.. 

So I propose the following. 
 - Complete the restructuring here. 
 - Modify the sync codes with current `ChainGraph` structure.
 - Merge this thing first.
 - Rebase #63 on top of it..
 
I haven't looked into #63 in details, but from there surface it seems some changes needs to be handled differently, weather it goes in before or after this PR. Namely we don't have a `esplora` crate anymore, and the idea was to use `rust-esplora` library directly and just use the client. So if we add extra APIs to esplora, the best place for that is `rust-esplora` not a lib here. Same for other things like electrum and rpc.

rajarshimaitra:
Updated the examples with the current version of the core API. 

Majorly the following changes
 - Removed all app related structs into `bdk_cli_example` lib. Including the argument match handler. 
 - `ChainGraph` is the main interface now instead of `TxGraph` and `SparseChain`.  Updating the structure is done by creating a new chain graph from blockchain data and applying over existing chain graph. 
 
NOTE on a `Trait`!!!

There's one new trait called `Broadcast`, but here me out, I promise it has a use. :)

In order to make the args handler function common for all examples (currently called `handle_commands`) it needs to handle broadcasting of transaction. But different backend have different api for broadcasting, so a trait is used to unify them.

I think this should be mostly done to show how different examples will fit with bdk_core.. 

Looking forward for suggestion and updates..

rajarshimaitra:
Theres conflict with the current master because of recent merge of #63 .. Not sure how to resolve them.. Suggestions welcome..

LLFourn:
Hey @rajarshimaitra. Unfortunately I think we have to de-prioritize merging this until after we have persistence working in #64 the existing example. From there this will be the top priority.

> There's one new trait called Broadcast, but hear me out, I promise it has a use. :)

I think having applications define traits is a fine idea. I'm not a fan of having `Broadcast` in bdk because we will have to have  a sync and async version and that's a mess.

rajarshimaitra:
> Hey @rajarshimaitra. Unfortunately I think we have to de-prioritize merging this until after we have persistence working in https://github.com/LLFourn/bdk_core_staging/pull/64 the existing example. From there this will be the top priority.

Awesome.. There will some simplification possible in the rpc example, if we have persistence.. So ACK on that going before this.. I will update it accordingly..  

> I think having applications define traits is a fine idea. I'm not a fan of having Broadcast in bdk because we will have to have a sync and async version and that's a mess.

Ya the scope of the `Broadcast` trait is only for the cli apps.. It doesn't have to be anything fundamental to bdk_core..

rajarshimaitra:
Updated over the current API. 

Summary of changes
 - `bdk_cli_lib` contains all the clap commands and command handlers
 - `bdk_esplora_example` uses `bdk_cli_lib` to create a cli app with esplora backend.
 -  local `esplora` APIs are removed. Using `rust-esplora-client::BlockingClient`. This requires one extra api `get_recent_blocks()` on the upstream. Which is done in this PR https://github.com/bitcoindevkit/rust-esplora-client/pull/29.. So `bdk_esplora_example` has my PR branch as dep for now.. 
 
 
 Upcoming: 
  -  Similar example crates for RPC and Electrum.. 
  
 Opened for early review and comments..

rajarshimaitra:
After creating the upstream PR i noticed theres already exists a similar PR https://github.com/bitcoindevkit/rust-esplora-client/pull/27.. 

@LLFourn would you prefer your PR instead for the `get_blocks()` API? That's definitely more details with a nice test too..

rajarshimaitra:
Pushed with a potential approach for #72  and some nits.. 

The `electrum.rs` looks very similar to the `esplora` module and has two functions `wallet_txid_scan` and `spk_txid_scan`.. The idea is electrum will sync in two stages.

 - `wallet_txid_scan` returns a candidate `KeychainScan` without TxGraph data.
 - The candidate is compared with tracker's `SparseChain`, and the resulting `ChangeSet` tells us all the new transactions we need to fetch..
 - New transactions are fetched in batch and added into the candidate `KeychainScan` update.
 - Apply the final `KeychainScan` on tracker and DB.
 
 In theory this should already handle conflicts and reorgs..
 
 For Reorg I tried a different approach this time..
 
  - candidate `KeychainScan`'s `Sparsechain` contains checkpoint of from `last_known_height - REORG_DEPTH` to `current_network_height`. This sparsechain when applied will handle reorg inconsistencies if any internally (in theory). So we don't need to have a `BTreeMap<u32, BlockHash>` in `wallet_txid_scan`, just an `Some(last_synced_height)`, or `None`, if tracker never synced before.. 
  
  I don't know if that's the right approach, but seemed to make the code simpler..

bdk_core/src/tx_graph.rs
```diff
@@ -51,8 +51,8 @@ impl TxGraph {
     }
 
     /// Returns true when graph contains given tx of txid (whether it be partial or full).
-    pub fn contains_txid(&self, txid: Txid) -> bool {
-        self.txs.contains_key(&txid)
+    pub fn contains_txid(&self, txid: &Txid) -> bool {
```
LLFourn, line 54: why change this?

bdk_esplora_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    let change_keychain = if let Some((internal_descriptor, internal_keymap)) = internal {
+        keymap.extend(internal_keymap);
+        keychain_tracker
+            .txout_index
+            .add_keychain(Keychain::Internal, internal_descriptor);
+        Keychain::Internal
+    } else {
+        Keychain::External
+    };
```
LLFourn, line 43: Everything from here above can go in the cli lib too.

LLFourn (commented):
Thanks @rajarshimaitra. Can we limit this PR to just esplora and have follow up PRs for rpc and electrum. This will mean we can merge the restructuring faster.

bdk_esplora_example/src/esplora.rs
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+            if let Some(last_active_index) = last_active_index {
+                wallet_scan
+                    .last_active_indexes
+                    .insert(keychain, last_active_index);
+            }
+        }
+
+        // This only works on last 10 blocks
```
LLFourn, line 165: This comment is only true if it's esplora backend. It might be 15 if it's mempool.space. I'd say either delete the comment or make this clear.

bdk_core/src/tx_graph.rs
rajarshimaitra: Because you only need a ref in the `hashMap::contains_key()`?? This was causing a move error somewhere downstream in the electrum example.. And required an extra cone, while the cloned object is dropped at this location only.. That was my reasoning behind this.. Open to change if this causes some other problems..

bdk_esplora_example/src/main.rs
rajarshimaitra: Hmm ya.. We can have a function like `init_trackers()`.. But maybe show in the example itself on how to create trackers and ppl might wanna build their own custom trackers in different examples?

bdk_esplora_example/src/esplora.rs
rajarshimaitra: Noted.. Will update..

rajarshimaitra:
> Can we limit this PR to just esplora and have follow up PRs for rpc and electrum.

I think I will open a new PR with restructure and esplora only.. Incorporate the comments, and turn this one into draft.. Maybe someone wanna check on the electrum and rpc sync too and have some comment on it, can use this one as a draft archive..

rajarshimaitra:
Converted to draft in favor of #73

LLFourn:
Superseded by #73

Issue #32, opened 2022-10-06
Restructure chain data indexing


evanlinjin:
Here are our current components:

1. SparseChain: It indexes transactions and where they are in blocks.
2. SpkTracker: It indexes txouts by script pubkeys in whatever transactions you add to it.

## Design goals

1. We want to be able to incrementally update `SpkTracker` with new data in the blockchain where complexity proportional to the changes (e.g. number of new transactions) since `SpkTracker` was last updated. This currently works but only in the case that new transactions are all later in the blockchain than when it was last updated.

2. We want to able to update the list of TXIDs in our TXID set in `SparseChain` without necessarily getting the full transaction data. Suggestion: split sprase chain into a (i) monotone structure which just stores and indexes transactions (but doesn't know which in the chain or not) and (ii) a non-monotone structure which keeps track of where transactions are in the chain but doesn't have full transaction data. We want to be able to quickly find which txids exist in the non-monotone structure but not in the monotone one.

3. `SpkTracker` currently has to scan over all the txouts to find which are unspent. We should be able to create an index of this. Preferably, this would be outside `SpkTracker`. Something like an `UnspentIndex`. `UnspentIndex` would be non-montone and would require the non-monotone blockchain data to generate.

## Potential Solution

1. Do not rely on the order of transactions in updates like the checkpointing system does. Rather, just cache the updates to each structure and then when other structures need to sync to it they can recall which was the last update they saw i.e. like git commits and merging.

## Potential Order

1. Purge codebase of internal checkpointing logic. `CheckPointCandidate` doesn't need to be changed here just the internal optimizing algorithms in `SpkTracker` and `SparseChain`.
2. Split `SparseChain` into monotone and non-monotone structures. Also change `CheckPointCandidate` to only have `txids` rather than full transactions.
3. Add commit like indexing from `SpkTracker` to monotone `SparseChain` and from monotone `SparseChain` to non-monotone `SparseChain`. (rename these monotone `SparseChain` could `TxGraph`).
4. Develop `UnspentIndex`.

LLFourn:
This looks good but actually I might swap 3 and 4 so APIs are solid first and then use commit structure to optimize.

evanlinjin:
TODO:

- [ ] Do we need marker type to enforce blocktime of tx?

evanlinjin:
# To Do (2)
- [ ] API to add auxiliary transactions to the mempool
[...]

PR #41, opened 2022-10-10
Remove checkpoint hash digests

evanlinjin:
Related to #32

### DIFF
--- a/bdk_core/src/sparse_chain.rs
+++ b/bdk_core/src/sparse_chain.rs
@@ -1,10 +1,7 @@
 use core::ops::RangeBounds;
 
 use crate::{collections::*, BlockId, BlockTime, Vec};
-use bitcoin::{
-    hashes::{sha256, Hash, HashEngine},
-    BlockHash, OutPoint, Script, Transaction, TxOut, Txid,
-};
+use bitcoin::{BlockHash, OutPoint, Script, Transaction, TxOut, Txid};
 
 #[derive(Clone, Debug, Default)]
 pub struct SparseChain {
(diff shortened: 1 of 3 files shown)


bdk_core/tests/test_sparse_chain.rs
```diff
@@ -149,7 +149,7 @@ fn checkpoint_limit_is_applied() {
     }
 
     assert_eq!(chain.iter_tx().count(), 10);
-    assert_eq!(chain.iter_checkpoints(..).count(), 5);
+    assert_eq!(chain.iter_checkpoints(..).count(), 10);
```
LLFourn, line 152: I would actually try and preserve the checkpoint limit here unless it really gets in the way.

bdk_core/src/spk_tracker.rs
```diff
@@ -52,31 +49,26 @@ impl<I: Clone + Ord> SpkTracker<I> {
...
-            .take_while(|checkpoint_id| {
-                !self
-                    .txid_digests_seen
-                    .contains(&chain.txid_digest_at(*checkpoint_id))
-            })
+            // .take_while(|checkpoint_id| {
+            //     !self
+            //         .txid_digests_seen
+            //         .contains(&chain.txid_digest_at(*checkpoint_id))
+            // })
```
LLFourn, line 56: delete rather than comment out this.

bdk_core/src/sparse_chain.rs
```diff
@@ -322,13 +275,6 @@ impl SparseChain {
                 deepest_change = Some(deepest_change.unwrap_or(u32::MAX).min(change));
             }
         }
-        if let Some(change_depth) = deepest_change {
```
LLFourn, line 325: Delete deepest_change variable that is unused now above.

LLFourn (approved):
LGTM.

1. I think we could keep the checkpoint limit (it will be there eventually).
2. Preferable to just delete stuff rather than comment out for review.

Merge when ready.

evanlinjin:
Probably will be replaced by #42

evanlinjin:
Definitely will be replaced by #42

PR #42, opened 2022-10-10
Introduce `TxGraph`

evanlinjin:
Progress on #32:
- [x] Purge codebase of internal checkpointing logic. `CheckPointCandidate` doesn't need to be changed here just the internal optimizing algorithms in `SpkTracker` and `SparseChain`.
- [x] Split `SparseChain` into monotone and non-monotone structures. Also change `CheckPointCandidate` to only have `txids` rather than full transactions.


bdk_core/tests/test_sparse_chain.rs
```diff
@@ -274,93 +222,74 @@ fn adding_checkpoint_where_tx_conftime_has_changed() {
...
-    );
+    let update3 = CheckpointCandidate {
+        base_tip: Some(BlockId {
+            height: 1,
+            hash: BlockHash::all_zeros(),
+        }),
+        invalidate: Some(update1.new_tip),
+        ..update2.clone()
+    };
+    assert_eq!(chain.apply_checkpoint(update3.clone()), ApplyResult::Ok);
```
LLFourn, line 269: This checkpoint doesn't make sense. it's invalidate is the same as it's base_tip. Should be an error.

LLFourn (changes requested):
I've done a quick first pass of `SparseChain` (haven't looked at `TxGraph`).

1. `SparseChain` is way simpler now. It's great. You should use that in tests. It's easy to generate a Vec of txids for each checkpoint. The update generating thing is too heavy a tool for that. It can probably be simplified a lot for `TxGraph` too.
2.  Some logic code deleted from `SparseChain` needs to be there.

bdk_core/tests/test_sparse_chain.rs
evanlinjin: I may have misunderstood either invalidate or base tip, but why does it need to be an error here?

Base tip is saying, we can only apply this update if the tip of sparsechain is this height and hash right?

And invalidate means, from this height and hash, all blocks and transactions should be removed?

bdk_core/src/sparse_chain.rs
```diff
@@ -121,103 +62,77 @@ impl SparseChain {
...
-            height,
-            hash: data.block_hash,
-        })
+        self.checkpoints
+            .get(&height)
+            .map(|&hash| BlockId { height, hash })
+    }
+
+    /// Return height of tx (if any).
+    pub fn transaction_at(&self, txid: &Txid) -> Option<Option<u32>> {
```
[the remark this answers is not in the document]
evanlinjin: We can make an enum? `{Height(u32)|Mempool}`

bdk_core/tests/test_sparse_chain.rs
LLFourn: > I may have misunderstood either invalidate or base tip, but why does it need to be an error here?
> 
> Base tip is saying, we can only apply this update if the tip of sparsechain is this height and hash right?

No. The fields mean:

- `base_tip` :  The **latest** checkpoint that is in the chain.
- `invalidate`: The **earliest** checkpoint that is not in the chain.

It might help to rename the fields to reflect this better.

bdk_core/src/sparse_chain.rs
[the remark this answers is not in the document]
LLFourn: I like this better.

bdk_core/src/tx_graph.rs
```diff
@@ -0,0 +1,83 @@
...
+            return false;
+        }
+
+        tx.input
+            .iter()
+            .map(|txin| txin.previous_output)
+            .for_each(|outpoint| {
+                self.spends
+                    .entry(outpoint)
+                    .or_insert_with(HashSet::new)
```
LLFourn, line 53: I think you can use `or_default` here too.

bdk_core/src/tx_graph.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        return true;
+    }
+
+    pub fn is_unspent(&self, outpoint: &OutPoint) -> Option<bool> {
+        self.spends.get(outpoint).map(|txids| txids.is_empty())
+    }
+
+    /// Return all txids of conflicting transactions to the given tx.
+    pub fn conflicting_txids(&self, tx: &Transaction, include_self: bool) -> Vec<Txid> {
```
LLFourn, line 71: suggestions:

- I don't think we ever want to `include_self`.
- Return an `impl Iterator`
- Make it an interator of `(OutPoint, u32)` meaning "the input at this `vin` of the passed in tx conflicts with the output at this `outpoint`.

bdk_core/src/spk_tracker.rs
```diff
@@ -116,19 +98,20 @@ impl<I: Clone + Ord> SpkTracker<I> {
...
 
     pub fn iter_txout<'a>(
         &'a self,
-        chain: &'a SparseChain,
+        graph: &'a TxGraph,
     ) -> impl DoubleEndedIterator<Item = (I, OutPoint)> + 'a {
         self.txouts
             .iter()
-            .filter(|(outpoint, _)| chain.get_tx(outpoint.txid).is_some())
+            .filter(|(outpoint, _)| graph.tx(&outpoint.txid).is_some())
```
LLFourn, line 114: I don't think this makes sense (it might not have made sense before). What this function used to do was to iterate over txouts that were in an actual chain. Now it iterates over txouts that have been added to `TxGraph`. Maybe change to `iter_txout_in_chain` and change the argument back to `SparseChain`.

bdk_core/src/spk_tracker.rs
```diff
@@ -39,51 +37,32 @@ impl<I> Default for SpkTracker<I> {
...
-
-        if let Some(latest_checkpoint) = chain.latest_checkpoint() {
-            let latest_txid_digest = chain.txid_digest_at(latest_checkpoint);
-            self.txid_digests_seen.insert(latest_txid_digest);
-        }
+    pub fn sync(&mut self, chain: &SparseChain, graph: &TxGraph) {
+        chain
+            .iter_txids()
+            .map(|(_, txid)| graph.tx(&txid).expect("tx must exist"))
+            .for_each(|tx| self.add_tx(tx));
```
LLFourn, line 49: I think we can just iter txs on the `graph` no need to pass in `chain`. Actually it would be good to be able to have a function with `iter_txout` on `graph` so when we add auxiliary txouts we can add them to the tracker without changing code here.

bdk_core/src/spk_tracker.rs
```diff
@@ -39,51 +37,32 @@ impl<I> Default for SpkTracker<I> {
...
     }
 
-    fn add_tx(&mut self, txid: Txid, chain: &SparseChain) {
-        let tx = &chain.get_tx(txid).expect("must exist").tx;
+    fn add_tx(&mut self, tx: &Transaction) {
```
LLFourn, line 52: I think this is better. It would be good to have `add_output(outpoint, txout)` too I think.

LLFourn (commented):
I reviewed the rest of the PR. Shaping up nicely :)

My main feeling is that we should take advantage of the simpler structure of these things in the tests. Probably trash most the checkpoint gen thing I did.

bdk_core/tests/test_sparse_chain.rs
evanlinjin: Let me know if my latest changes make sense.

bdk_core/src/tx_graph.rs
evanlinjin: Do you mean an iterator of `(u32, Txid)`?

bdk_core/src/spk_tracker.rs
evanlinjin: Does mempool txouts go here as well?

bdk_core/src/spk_tracker.rs
evanlinjin: Done.

bdk_core/src/spk_tracker.rs
evanlinjin: Done.

evanlinjin:
Adding `PartialTx` in `TxGraph` now.

bdk_core/src/sparse_chain.rs
```diff
@@ -226,447 +139,229 @@ impl SparseChain {
...
-                hasher.input(&txid);
-            }
-            let txid_hash = sha256::Hash::from_engine(hasher);
-            self.txid_digest_at(tip) == txid_hash
-        } else {
-            true
-        }
+    pub fn prune_checkpoints(&mut self) -> Option<BTreeMap<u32, BlockHash>> {
+        let height = *self.checkpoints.keys().rev().nth(self.checkpoint_limit?)?;
+        return Some(self.checkpoints.split_off(&height));
```
LLFourn, line 317: This logic seems wrong. If my checkpoint limit is `Some(0)` then it should drain all the checkpoints (0 checkpoints are allowed). If it's `Some(1)` only the most recent one should be kept. 

`split_off` is splitting the more recent ones though :S

bdk_core/tests/test_sparse_chain.rs
```diff
@@ -132,11 +68,13 @@ fn out_of_order_tx_is_before_first_checkpoint() {
 fn checkpoint_limit_is_applied() {
     let mut checkpoint_gen = CheckpointGen::new();
     let mut chain = SparseChain::default();
-    chain.set_checkpoint_limit(5);
+    let mut graph = TxGraph::default();
+    // chain.set_checkpoint_limit(5);
```
LLFourn, line 72: :thinking: :thinking:

LLFourn (changes requested):
Getting closer :)

I hope what I am proposing  makes sense to you with respect to `last_valid`, `invalidate` and `new_tip`. Please make sure it does. It might be good try and re-write the docs for them to see if you can articulate the logic.

At risk of repeating myself,I think you need to test each component individually. The checkpointgen thing is not useful anymore and makes the tests harder to understand.

Let's delete every API in `SpkTracker` that uses `SparseChain` for now. It should only interact with `TxGraph`. This means we have to do a little bit of filtering in the example application but that's ok. It's worth it to get rid of the testing surface.

bdk_core/src/spk_tracker.rs
```diff
@@ -116,22 +89,35 @@ impl<I: Clone + Ord> SpkTracker<I> {
     pub fn iter_txout_full<'a>(
         &'a self,
         chain: &'a SparseChain,
+        graph: &'a TxGraph,
     ) -> impl DoubleEndedIterator<Item = (I, FullTxOut)> + 'a {
         self.txouts.iter().filter_map(|(outpoint, spk_index)| {
-            Some((spk_index.clone(), chain.full_txout(*outpoint)?))
+            Some((spk_index.clone(), chain.full_txout(graph, *outpoint)?))
         })
```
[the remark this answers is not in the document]
evanlinjin: So should I just remove `iter_txout_full` completely?

bdk_core/src/sparse_chain.rs
evanlinjin: Should be fixed, need to test.

bdk_core/tests/test_sparse_chain.rs
evanlinjin: Added back

bdk_core/src/spk_tracker.rs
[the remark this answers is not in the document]
LLFourn: I think `iter_txout_full` is ok since it only needs `graph`? But I'm fine with deleting helper type APIs atm.

bdk_core/src/sparse_chain.rs
```diff
@@ -226,447 +139,252 @@ impl SparseChain {
...
-            });
-
-        let mut deepest_change = None;
-        for tx in new_checkpoint.transactions {
-            if let Some(change) = self.add_tx(tx) {
-                deepest_change = Some(deepest_change.unwrap_or(u32::MAX).min(change));
+            // ensure next checkpoint is the same as `invalidate`
+            if !matches!(next_block.as_ref(), Some(b) if b == invalidate) {
+                return ApplyResult::Stale(StaleReason::InvalidateDoesNotPreceedLastValid {
+                    preceeds_last_valid: next_block,
```
LLFourn, line 184: succeeds?

LLFourn (commented):
Just had a look at most recent changes. Looks great :)

bdk_core/src/sparse_chain.rs
```diff
@@ -226,447 +139,252 @@ impl SparseChain {
...
-                    return ApplyResult::Stale(StaleReason::InvalidationHashNotMatching {
-                        invalidates: checkpoint_reset,
-                        expected: None,
-                    })
-                }
+            // new tip should be of equal or greater height than last_valid
+            if last_valid.height > new_checkpoint.new_tip.height
+                // if new tip is at the same height as last valid, block hashes should also match
+                || last_valid.height == new_checkpoint.new_tip.height
+                    && last_valid.hash != new_checkpoint.new_tip.hash
```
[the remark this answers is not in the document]
LLFourn: I made a mistake. After looking at this I think it makes things worse :flushed:. Now we have to deal with `new_tip` and `last_valid` being `None`. It also is more implicit which means we catch less mistakes.

bdk_core_example/src/main.rs
```diff
@@ -189,29 +190,37 @@ fn main() -> anyhow::Result<()> {
...
 
             println!("confirmed: {}", confirmed);
             println!("unconfirmed: {}", unconfirmed);
         }
         Commands::Txo { utxo_cmd } => match utxo_cmd {
             TxoCmd::List => {
-                for (spk_index, txout) in tracker.iter_txout_full(&chain) {
+                for (spk_index, txout) in
+                    tracker
+                        .iter_unspent(&chain, &graph)
```
LLFourn, line 213: This now only iters unspent. It's meant to iter all txouts in the chain. Should be `iter_txout` instead.

bdk_core/tests/test_sparse_chain.rs
```diff
@@ -1,529 +1,361 @@
...
-    update.transactions[0].confirmation_time = Some(BlockTime { height: 1, time: 1 });
-    assert!(matches!(
-        chain.apply_checkpoint(update.clone()),
-        ApplyResult::Inconsistent { .. }
-    ));
-    assert_eq!(chain.iter_checkpoints(..).count(), 1);
+    assert_eq!(chain.iter_confirmed_txids().count(), 10);
+    let latest = chain.latest_checkpoint();
+    assert_eq!(chain.iter_checkpoints(..).count(), 5);
+    assert_eq!(chain.latest_checkpoint(), latest);
```
LLFourn, line 203: I don't think you meant to check the variable you assigned two lines above.

LLFourn (approved):
Great work. A few minor changes suggested.

+ can you remove most of the TODOs here and just make issues out of the important them saying exactly what we have to do to address. Thanks!

Once the above done, merge when you're happy.

bdk_core/tests/test_sparse_chain.rs
evanlinjin: What was I thinking?

PR #45, opened 2022-10-14
Update varint_size to use rust-bitcoin values

danielabrozzoni:
Sorry for the nit 🥺


evanlinjin (commented):
utACK

Brownie points for the test. However @LLFourn is doing some massive refactoring with the coin select module, which may cause conflicts.

danielabrozzoni:
No problem for me, it can stay open for months. The current code is good anyways

LLFourn (approved):
Thanks! I think I might remove the tests before merging since if there's tests for this they should be in rust-bitcoin.

danielabrozzoni:
> Thanks! I think I might remove the tests before merging since if there's tests for this they should be in rust-bitcoin.

Sure, I put them there just to double check that it behaved exactly like the old implementation

danielabrozzoni:
Removed :)

PR #46, opened 2022-10-18
Simplify coin selector and branch and bound

LLFourn:
1. `CoinSelector` now only takes a `base_weight` parameter. It doesn't store your `target_value`, `feerate` etc. You can apply policies on these outside.
2. You can sort the candidates inside `CoinSelector` so that for calls to `unselected` they will appear in that order without changing the original order.
3. You can ban candidates in the coin selector so that they won't appear in `unselected` anymore.
4. "branch and bound" is implemented generically :tada: -- you can pass in your own scoring function. The scoring function is passed the current selection and a bool to indicate whether it should return the score of the current selection or the lower bound score for any child selection of the current one (i.e. the best it could do by adding more inputs). Note textbook branch and bound would have two separate functions but because of difficulties of rust closures they had to be one.
5. I show how to minimize waste metric on top of branch and bound implementation.

This is a WIP but I think that it's good enough for me to explore how (1) affects the API in BDK.

Early impressions appreciated.

### DIFF
--- a/bdk_coin_selection/tests/waste.proptest-regressions
+++ b/bdk_coin_selection/tests/waste.proptest-regressions
@@ -1,0 +1,10 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc b526e3a05e5dffce95e0cf357f68d6819b5b92a1c4abd79fd8fe0e2582521352 # shrinks to num_inputs = 45, target = 16494, feerate = 3.0291684, min_fee = 0, base_weight = 155, long_term_feerate_diff = -0.70271873, change_weight = 58, change_spend_weight = 82
+cc f3c37a516004e7eda9183816d72bede9084ce678830d6582f2d63306f618adee # shrinks to num_inputs = 40, target = 6598, feerate = 8.487553, min_fee = 221, base_weight = 126, long_term_feerate_diff = 3.3214626, change_weight = 18, change_spend_weight = 18
+cc a6d03a6d93eb8d5a082d69a3d1677695377823acafe3dba954ac86519accf152 # shrinks to num_inputs = 49, target = 2917, feerate = 9.786607, min_fee = 0, base_weight = 4, long_term_feerate_diff = -0.75053596, change_weight = 77, change_spend_weight = 81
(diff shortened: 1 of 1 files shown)


bdk_core/src/coin_select/coin_selector.rs
```diff
@@ -40,124 +42,85 @@ impl WeightedValue {
...
+    pub fn apply_selection<T>(&self, candidates: &'a [T]) -> impl Iterator<Item = &'a T> + '_ {
+        self.selected.iter().map(|i| &candidates[*i])
+    }
 
-impl<'a> CoinSelector<'a> {
-    pub fn candidate(&self, index: usize) -> &WeightedValue {
-        &self.candidates[index]
+    pub fn select(&mut self, index: usize) -> bool {
+        assert!(index < self.candidates.len());
+        self.selected.to_mut().insert(index)
```
rajarshimaitra, line 105: I don't know if I am reading wrong, but it seems the `index` here is the `index of the item in the candidate list`?? Where as the value of the stuff in that index is the actual index of the actual utxo (or whatever we are selecting)?? If so, Why keep so many indexes? Cant we just directly select the utxo index in the selected list?

rajarshimaitra (commented):
I just made a first pass over the new bnb code, and I won't pretend I understood it fully, but it does look structurally very different from the existing bnb or any other bnb codes I have seen so far.. 

I "kinda" get what its trying to do inside the `BnbIter`.. But I am happy that I am now finally motivated to go through Murch's thesis to weigh on this code.. 😅 

One generic question I had.. Does a "generic branch and bound" means we can use the same algorithm to optimize for different "policies"?? If so, what kind of different policies are we talking about?? Is it (theoretically) possible to combine various existing CS algos into this generic model?? 

Will make another thorough pass after this.. 

Below are some more questions and comments..

LLFourn:
Update: I've moved the coin selection to its own crate that only depends on `rust-bitcoin` as requested in https://github.com/revault/minisafe/issues/51#issuecomment-1280446716. It's also easier like this to start putting it into `bdk`.

I've tried to design branch and bound algorithms so they can be composed e.g. if you want to find a "changeless" solution but then tiebreak by the waste metric i hoped you could easily create a scoring function that returns `(has_change, waste_metric(..))`. Since this is `Ord` it should "just work". However, this is subtly very tricky. That works for the *score* of a particular selection but not the *heuristic* (or the "bound"). How can we get the lower bound of the boolean indicating whether change was used? You could always return `false` which would work but it would be worse than if we created a combined heuristic targeted to do the job properly.

@rajarshimaitra thanks for the initial review. I'll try and respond to comments tomorrow.

PR #47, opened 2022-10-18
Use the new planning module

afilini:


evanlinjin:
Just a note, some of the tests might be just wrong.

LLFourn:
Thanks @afilini. Where this would be most useful is in the `bdk_core_example` crate rather than `bdk_core`. Sightly off topic but I'm actually playing with the idea that `bdk_core` shouldn't even depend on miniscript at all (only `rust-bitcoin`). Then the composition of the primitives in `bdk_core` with miniscript could be done in `bdk` which gives a cleaner separation of concerns. Of course this assumes that the planning module will end up in `miniscript`.

I will take a look at the planning code soon.

PR #49, opened 2022-10-21
Implement `UnspentIndex` and `ChangeSet`

evanlinjin:
* `SparseChain::apply_update` now returns a `ChangeSet` of all altered `txid`s and `checkpoint`s.
* `SpkTracker::sync` now uses both `TxGraph` and `SparseChain`, to only iterate over relevant txs.
* Implement `TxGraph::calculate_fee` which calculates fee of a given tx.
* Implement `UnspentIndex`.
* Added tests and various testing improvements.


bdk_core/src/sparse_chain.rs
```diff
@@ -242,63 +261,170 @@ impl SparseChain {
...
+            txids: [(txid, height)].into(),
+            ..Update::new(Some(tip), tip)
+        };
+
+        self.apply_update(update).map_err(|err| Some(err))
+    }
+
+    /// Insert an arbitary txid into the mempool. Returns [`None`] when tx already exists in
+    /// [`SparseChain`], otherwise it returns the [`ChangeSet`].
+    pub fn insert_mempool_tx(&mut self, txid: Txid) -> Option<ChangeSet> {
```
LLFourn, line 370:
```suggestion
    pub fn insert_unconfirmed_tx(&mut self, txid: Txid) -> Option<ChangeSet> {
```

LLFourn (changes requested):
@evanlinjin how about we make these two changes separate PRs?

`UnspentIndex` can have a `sync` method that takes `sparse_chain`, `tx_graph` and `spk_tracker` for now and we can defer designing and implementing "changeset" stuff.  I think we both agree that doing changsets should commit to each previous one for the sake of easily detecting whether the changeset can actually be applied. I like what you've done and I think it's worth getting it right in its own PR rather than merging a half-finished idea now.

Once we figure that out we can use chagesets in `sync` behind the scenes to make it go faster and also expose them for inspection.

PR #50, opened 2022-10-25
Switch to codecov-action v3

danielabrozzoni:
- Switch to codecov-action v3, as v1 has been deprecated in February 2022, and no longer works.
- Upload the HTML report to the CI artifacts - so that if codecov doesn't work for some reason, we can still inspect the report.

### DIFF
--- a/.github/workflows/codecov.yaml
+++ b/.github/workflows/codecov.yaml
@@ -31,7 +31,7 @@ jobs:
       
       # Upload coverage report
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v1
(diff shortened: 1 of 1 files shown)


danielabrozzoni:
I'm still getting an error when uploading to codecov, and https://github.com/codecov/codecov-action/issues/295 seems to suggest that we need to manually grant access to the repo. @LLFourn can you do it, please? I think you're the only one who can, as you're the repository owner.

LLFourn:
I think I did this now. Sorry I kept trying to do this but not figuring it out and getting distracted. I'm a little concerned that it needed permission to write all my public repos just to do this one.

PR #53, opened 2022-11-01
Graph extend

LLFourn:
Needed to extend the graph with another easily so that I can gather new transactions into a graph, scan them with the spk tracker and then add them to the existing graph.

### DIFF
--- a/bdk_core/src/tx_graph.rs
+++ b/bdk_core/src/tx_graph.rs
@@ -41,14 +41,19 @@ impl TxGraph {
     pub fn txout(&self, outpoint: OutPoint) -> Option<&TxOut> {
         match self.txs.get(&outpoint.txid)? {
             TxNode::Whole(tx) => tx.output.get(outpoint.vout as usize),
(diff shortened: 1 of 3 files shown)


bdk_core/src/tx_graph.rs
```diff
@@ -127,6 +132,21 @@ impl TxGraph {
...
+            TxNode::Partial(_) => None,
+        })
+    }
+
+    pub fn iter_partial_transactions(&self) -> impl Iterator<Item = (Txid, &BTreeMap<u32, TxOut>)> {
+        self.txs.iter().filter_map(|(txid, tx)| match tx {
+            TxNode::Whole(_) => None,
+            TxNode::Partial(partial) => Some((*txid, partial)),
+        })
+    }
```
LLFourn, line 148: Note I don't think I actually need these right now but added them since it feels like that's a basic thing you might want to do.

evanlinjin (approved):
ACK  9d20dfbf23655353fc24cd8272b7a076ea19690c Looks good to me.

LLFourn:
Merging despite code coverage thingo since we're in the middle of testing out APIs.

PR #60, opened 2022-11-08
Introduce `SparseChain` index extension

evanlinjin:
> This description is mostly copied from #61

Usually transactions are indexed by their height. But you might want to extend this (e.g. include block position) or some other metadata.

I've made the index quite restrictive (e.g. requires Copy) since we'll need to persist the data and for our simple file based persistence we'll need it to be fixed size for now.

This fixes a bug where an inclusive txid range search would not find the txids because we set the range to `Txid::all_zeros` (unless the txid was all zeroes). 

I also added an API to search via the index itself. Handy if you want to do pagination of  transactions by your own index (a good one would be block position) as mentioned in https://github.com/bitcoindevkit/bdk/issues/794

### To Do
- [x] Add test for *"This fixes a bug where an inclusive txid range search would not find the txids because we set the range to `Txid::all_zeros` (unless the txid was all zeroes)"*.

### DIFF
--- a/bdk_core/src/sparse_chain.rs
+++ b/bdk_core/src/sparse_chain.rs
@@ -6,16 +6,27 @@ use core::{
 use crate::{collections::*, BlockId, TxGraph, Vec};
 use bitcoin::{hashes::Hash, BlockHash, OutPoint, Transaction, TxOut, Txid};
 
-#[derive(Clone, Debug, Default)]
-pub struct SparseChain<D = ()> {
+#[derive(Clone, Debug)]
+pub struct SparseChain<E = ()> {
     /// Block height to checkpoint data.
     checkpoints: BTreeMap<u32, BlockHash>,
     /// Txids prepended by confirmation height.
-    txid_by_height: BTreeSet<(TxData<D>, Txid)>,
+    indexed_txids: BTreeSet<(ChainIndex<E>, Txid)>,
     /// Confirmation heights of txids.
-    txid_to_index: HashMap<Txid, TxData<D>>,
+    txid_to_index: HashMap<Txid, ChainIndex<E>>,
     /// Limit number of checkpoints.
     checkpoint_limit: Option<usize>,
+}
+
+impl<E> Default for SparseChain<E> {
+    fn default() -> Self {
+        Self {
+            checkpoints: Default::default(),
+            indexed_txids: Default::default(),
+            txid_to_index: Default::default(),
+            checkpoint_limit: Default::default(),
+        }
+    }
 }
 
 #[derive(Clone, Debug, PartialEq)]
@@ -79,7 +90,7 @@ impl core::fmt::Display for UpdateFailure {
 #[cfg(feature = "std")]
 impl std::error::Error for UpdateFailure {}
 
-impl<D: Clone + Debug + Default + Ord> SparseChain<D> {
+impl<E: ChainIndexExtension> SparseChain<E> {
     /// Creates a new chain from a list of blocks. The caller must guarantee they are in the same
     /// chain.
     pub fn from_checkpoints(checkpoints: impl IntoIterator<Item = BlockId>) -> Self {
@@ -105,8 +116,8 @@ impl<D: Clone + Debug + Default + Ord> SparseChain<D> {
             .map(|&hash| BlockId { height, hash })
     }
 
-    /// Return the associated data of a tx of txid (if any).
-    pub fn tx_data(&self, txid: Txid) -> Option<TxData<D>> {
+    /// Return the associated index of a tx of txid (if any).
+    pub fn tx_index(&self, txid: Txid) -> Option<ChainIndex<E>> {
         self.txid_to_index.get(&txid).cloned()
     }
 
@@ -126,7 +137,7 @@ impl<D: Clone + Debug + Default + Ord> SparseChain<D> {
     }
 
     /// Derives a [`ChangeSet`] that could be applied to an empty index.
-    pub fn initial_change_set(&self) -> ChangeSet<D> {
+    pub fn initial_change_set(&self) -> ChangeSet<E> {
         ChangeSet {
             checkpoints: self
                 .checkpoints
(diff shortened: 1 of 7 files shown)


LLFourn (commented):
Leaving some comments. Not sure if what i suggests makes sense. Thinking about it some more.

bdk_core/src/sparse_chain.rs
```diff
@@ -350,27 +352,27 @@ impl SparseChain {
...
         range: R,
     ) -> impl DoubleEndedIterator + '_ {
-        fn map_bound(bound: Bound<&TxHeight>) -> Bound<(TxHeight, Txid)> {
+        fn map_bound<D: Clone + Default>(bound: Bound<&TxHeight>) -> Bound<(TxData<D>, Txid)> {
             match bound {
                 Bound::Unbounded => Bound::Unbounded,
-                Bound::Included(x) => Bound::Included((x.clone(), Txid::all_zeros())),
-                Bound::Excluded(x) => Bound::Excluded((x.clone(), Txid::all_zeros())),
+                Bound::Included(x) => Bound::Included((x.clone().into(), Txid::all_zeros())),
+                Bound::Excluded(x) => Bound::Excluded((x.clone().into(), Txid::all_zeros())),
```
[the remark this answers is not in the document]
evanlinjin: Good point, how about:

```rust
pub struct TxData<D> {
    pub height: TxHeight,
    pub data: Option<D>,
}
```

This way when `data` is none, it will always be smaller than when data is `Some`.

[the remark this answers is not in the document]
LLFourn: I'm not sure that solves the problem. You still need to find the `max` of `Option<D>` to put it in the bounds. I think you want to people to be able to query on the `D`. I mean that's why you've put it in this index right?

We have to decide whether the data is meant to be changing the order of stuff (in which case it should be included in the range) or whether it's just additional data in which case we don't put it in the height index.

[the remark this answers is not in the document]
evanlinjin: I feel like querying by `TxHeight` should be enough for now? As long as we have `min` for `Option<D>` we can query by `TxHeight` only.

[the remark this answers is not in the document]
evanlinjin: Fixed here:

https://github.com/LLFourn/bdk_core_staging/blob/430c2cf7478663cabf5c2d984c5c16b5fdda51a7/bdk_core/src/sparse_chain.rs#L392-L424

bdk_core/src/sparse_chain.rs
```diff
@@ -638,23 +652,54 @@ impl Display for MergeFailure {
...
+    const MAX: Self = u32::MAX;
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ChainIndex<E = ()> {
+    pub height: TxHeight,
+    pub extension: E,
+}
+
+impl<E: Display> Display for ChainIndex<E> {
```
LLFourn, line 680: I think `ChainIndex` only needs `Debug`.

bdk_esplora/src/ureq.rs
```diff
@@ -182,7 +182,7 @@ impl Client {
         mut scripts: impl Iterator<Item = (u32, Script)> + Clone,
         stop_gap: usize,
         existing_chain: &BTreeMap<u32, BlockHash>,
-    ) -> Result<(Option<u32>, ChainGraph), UpdateError> {
+    ) -> Result<(Option<u32>, ChainGraph<()>), UpdateError> {
```
LLFourn, line 185: Isn't `()` the default?

bdk_core/src/sparse_chain.rs
```diff
@@ -350,27 +362,75 @@ impl SparseChain {
...
+        self.indexed_txids.range((
+            map_bound(
+                range.start_bound(),
+                (E::MIN, min_txid()),
+                (E::MAX, max_txid()),
+            ),
+            map_bound(
+                range.end_bound(),
+                (E::MAX, max_txid()),
+                (E::MIN, min_txid()),
```
LLFourn, line 427: riiight very subtle logic there. Nice work.

bdk_core/src/sparse_chain.rs
```diff
@@ -180,15 +185,17 @@ impl SparseChain {
...
 
-        for (tx_height, txid) in &update.txid_by_height {
+        for (tx_index, txid) in &update.indexed_txids {
             // ensure all currently confirmed txs are still at the same height (unless, if they are
             // to be invalidated, or originally unconfirmed)
-            if let Some(&old_height) = self.txid_to_index.get(txid) {
-                if old_height < TxHeight::Confirmed(invalid_from) && tx_height != &old_height {
+            if let Some(old_tx_index) = self.txid_to_index.get(txid) {
+                if old_tx_index.height < TxHeight::Confirmed(invalid_from)
+                    && tx_index.height != old_tx_index.height
```
LLFourn, line 193: This should be comparing the index. You can't change the index either without a reorg (e.g. position in block, or timestamp etc).

bdk_core/src/sparse_chain.rs
```diff
@@ -180,15 +185,17 @@ impl SparseChain {
...
+                {
                     return Err(UpdateFailure::InconsistentTx {
                         inconsistent_txid: *txid,
-                        original_height: old_height,
-                        update_height: *tx_height,
+                        original_height: old_tx_index.height,
+                        update_height: tx_index.height,
```
LLFourn, line 198: The failure should contain the index.

bdk_core/src/sparse_chain.rs
```diff
@@ -308,28 +315,33 @@ impl SparseChain {
...
 
-        if height.is_confirmed() && (latest.is_none() || height > latest.unwrap()) {
+        if index.height.is_confirmed() && (latest.is_none() || index.height > latest.unwrap()) {
             return Err(InsertTxErr::TxTooHigh);
         }
 
-        if let Some(&old_height) = self.txid_to_index.get(&txid) {
-            if old_height.is_confirmed() && old_height != height {
+        if let Some(old_index) = self.txid_to_index.get(&txid) {
+            if old_index.height.is_confirmed() && old_index.height != index.height {
```
LLFourn, line 336: same here.

LLFourn (changes requested):
Nice. This is better. Needs some tests against changing the "extension" but not the height. That should be an error (without a reorg) -- unless it's unconfirmed of course. We should also probably have a way of telling the thing that the extension creates a total order so nothing should exist at the same spot unless it's equal e.g. two txids cannot have the same position in the same block. But we can leave that for later.

PR #63, opened 2022-11-10
Introduce `range_txids_by_height` and introduce timestamps to esplora example

evanlinjin:
Changes:
* Introduce `ConfirmationTime` which replaces `BlockTime`.
* `ChainIndex` is now a trait (which `TxHeight` and `ConfirmationTime` implements.
* Add `SparseChain::range_txids_by_height()` with tests.
* Add comments and documentation.
* Esplora example now stores timestamp in `SparseChain`.

### DIFF
--- a/bdk_core/src/lib.rs
+++ b/bdk_core/src/lib.rs
@@ -1,16 +1,15 @@
 #![no_std]
 pub use alloc::{boxed::Box, vec::Vec};
 pub use bitcoin;
-use bitcoin::{hashes::Hash, BlockHash, TxOut};
 mod chain_graph;
 pub use chain_graph::*;
 mod spk_tracker;
 pub use spk_tracker::*;
-mod sparse_chain;
-pub use sparse_chain::*;
-mod tx_graph;
-pub use tx_graph::*;
+mod chain_data;
 pub mod coin_select;
+pub mod sparse_chain;
+pub mod tx_graph;
+pub use chain_data::*;
 #[cfg(feature = "miniscript")]
 mod keychain_tracker;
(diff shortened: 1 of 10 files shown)


bdk_core/src/lib.rs
```diff
@@ -61,121 +60,8 @@ pub mod collections {
...
-#[derive(Clone, Debug, PartialEq)]
-pub enum PrevOuts {
-    Coinbase,
-    Spend(Vec<TxOut>),
-}
+// #[derive(Clone, Debug, PartialEq)]
+// pub enum PrevOuts {
+//     Coinbase,
+//     Spend(Vec<TxOut>),
+// }
```
evanlinjin, line 67: Should this be removed?

bdk_core/src/sparse_chain.rs
```diff
@@ -629,107 +672,67 @@ impl<E: core::fmt::Debug> Display for MergeFailure<E> {
...
         }
     }
-}
 
-impl TxHeight {
-    pub fn is_confirmed(&self) -> bool {
-        matches!(self, Self::Confirmed(_))
+    #[test]
+    fn verify_tx_height() {
+        verify_chain_index::<TxHeight>(1000, 1000);
```
LLFourn, line 738: I guess do this for confirmation time too? I think it would be better to put this test in `chain_data.rs` because that's where the impls are.

LLFourn (changes requested):
This looks great. A few things to clean up.

I pushed one change to clean up `from_checkpoints` API to stop you needing to specify type parameters.

PR #64, opened 2022-11-11
Add persistance with example database

vladimirfomene:


bdk_core/src/byte_list.rs
```diff
@@ -95,19 +99,20 @@ impl<F: Read + Write + Seek> DiskList<F> {
         header[0..8].copy_from_slice(tail_loc.to_be_bytes().as_ref());
         header[8..12].copy_from_slice((data.len() as u32).to_be_bytes().as_ref());
         // Write new entry first so we're sure it's written before we update pointer to tail
-        self.inner.write_all(&header[..])?;
+        self.inner.write_all(&header)?;
         self.inner.write_all(data)?;
         self.inner.flush()?;
         // it worked so let's set the pointer
         self.set_tail_loc(key, new_tail_loc)?;
-        self.inner.flush()?;
```
LLFourn, line 103: :+1:

bdk_core/src/byte_list.rs
```diff
@@ -2,13 +2,13 @@ use alloc::vec::Vec;
 use core::mem::size_of;
 use std::io::{self, Read, Seek, SeekFrom, Write};
 
-/// A very simple collection of `256` on disk linked lists.
+/// A very simple collection of `256` byte linked lists.
 #[derive(Debug)]
-pub struct DiskList<F> {
-    inner: F,
+pub struct ByteList<S> {
```
LLFourn, line 7: It's not a list of bytes :). I suppose it's not a list of disks either.

Just call it `ExampleDb` or something. It probably shouldn't leave this repo.

LLFourn (commented):
Thanks. Let's just do the renaming I suggested. It should be clear it's just here as an example of the simplest possible database API that you need to use BDK.

bdk_core/src/byte_list.rs
```diff
@@ -95,19 +99,20 @@ impl<F: Read + Write + Seek> DiskList<F> {
         header[0..8].copy_from_slice(tail_loc.to_be_bytes().as_ref());
         header[8..12].copy_from_slice((data.len() as u32).to_be_bytes().as_ref());
         // Write new entry first so we're sure it's written before we update pointer to tail
-        self.inner.write_all(&header[..])?;
+        self.inner.write_all(&header)?;
         self.inner.write_all(data)?;
         self.inner.flush()?;
```
LLFourn, line 104: I did some digging and it turns out that this is wrong. We should be using `File` type here not `Read + Write + Seek`: https://doc.rust-lang.org/src/std/sys/unix/fs.rs.html#1061

Instead of `.flush` we should call `.datasync`.

vladimirfomene:
@LLFourn, I have updated as recommended and finished implementing the sync functionality. Couldn't get the first test to pass, it is timing for some reason? I believe it is become of the number of file i/o that the proptest is trying to do.

bdk_core_example/src/main.rs
```diff
@@ -245,14 +245,19 @@ fn main() -> anyhow::Result<()> {
...
-            });
+                spks.extend(tracker.iter_unused().into_iter().map(|(index, script)| {
+                    eprintln!("scanning {:?}", index);
+                    script.clone()
+                }));
+            } else {
+                spks.extend(tracker.script_pubkeys().into_iter().map(|(index, script)| {
+                    eprintln!("scanning {:?}", index);
+                    script.clone()
+                }));
```
LLFourn, line 256: There should be two scanning options: `unused` and `unspent`. `unspent` should add the spks of all unspent utxos to the list. If nothing is specified just assume `unused=true`, and `unspent=true`. We can add an `--all` option to scan all spks given out too (what you implemented) if we want as well but that should definitely be opt in.

LLFourn (changes requested):
I hope the what I've suggested makes sense :)

LLFourn:
I changed the PR's base to master. There will be a lot of conflicts as we go but we might as well use this PR to do all the changes.

bdk_core_example/src/main.rs
vladimirfomene: Okay got it

bdk_core/src/example_db.rs
```diff
@@ -57,15 +60,17 @@ macro_rules! read_ints {
     }}
 }
 
-impl<F: Read + Write + Seek> DiskList<F> {
-    pub fn load(mut file: F) -> Result<Self, LoadError> {
+impl ExampleDB {
+    /// Load file ready for Read/Write/Seek
+    pub fn load(mut file: File) -> Result<Self, LoadError> {
```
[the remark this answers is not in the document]
vladimirfomene: Okay got it

bdk_core/src/sparse_chain.rs
```diff
@@ -247,43 +242,35 @@ impl SparseChain {
...
-                    self.txid_to_index.insert(txid, to),
-                ),
-            };
-            debug_assert!(changed);
-            debug_assert_eq!(original_height, change.from);
+        for (&txid, &new_index) in &changeset.txids {
+            match new_index {
+                Some(new_index) => {
+                    self.txid_by_height.insert((new_index, txid));
+                    self.txid_to_index.insert(txid, new_index);
```
evanlinjin, line 263: We need to remove the old entry of `txid`.

evanlinjin:
I'm going to create a separate PR to have the `ChangeSet`-related changes only.

bdk_core/src/tx_graph.rs
```diff
@@ -168,24 +168,72 @@ impl TxGraph {
...
+
+#[derive(Debug, Clone, PartialEq, Default)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Deserialize, serde::Serialize),
+    serde(crate = "serde_crate")
+)]
+pub struct Additions {
+    pub tx: BTreeSet<Transaction>,
+    pub txout: BTreeMap<OutPoint, TxOut>,
```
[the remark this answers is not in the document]
LLFourn: `TxNode` is an internal type and should stay that way IMO.

bdk_core_example/src/main.rs
LLFourn: What you just pushed looks good but `--all` necessarily includes all the `spks` from unspent and unsued so there's no need to add those explicitly if `--all` is set.

bdk_core_example/src/main.rs
```diff
@@ -240,21 +244,36 @@ fn main() -> anyhow::Result<()> {
...
                     script.clone()
                 }));
-            };
+            }
+
+            if unspent {
+                spks.extend(tracker.iter_unspent(chain.chain(), chain.graph()).map(
+                    |(index, ftxout)| {
+                        eprintln!("scanning {:?}", index);
+                        ftxout.txout.script_pubkey
```
LLFourn, line 273: what about:

```rust
eprintln!("checking if {} has been spent", ftxout.outpoint);
```

vladimirfomene:
Hi @LLFourn ! I have updated the PR as recommended. I was wondering if you can help me resolve the conflicts, I'm finding it hard to tell in some cases which is suppose to be most recent code.

vladimirfomene:
Hi @LLFourn ! I have updated the PR as recommended. I was wondering if you can help me resolve the conflicts, I'm finding it hard to tell in some cases which is suppose to be most recent code.

bdk_core_example/src/main.rs
vladimirfomene: By default now, it will scan all and only scan `unused` or `unspent` when explicitly requested

vladimirfomene:
Better implementation merged in #67

PR #66, opened 2022-11-16
Use the new planning module

afilini:
Use the new planning module in the example

This PR adds a bit of new code because our version of the planning module in miniscript is a bit simpler, so we have to do more things here.

This also starts using PSBTs as a storage medium for signatures, to remove the need for a special "signatures container" struct which was present before.

### DIFF
--- a/bdk_core_example/src/main.rs
+++ b/bdk_core_example/src/main.rs
@@ -3,13 +3,17 @@ use bdk_core::{
     bitcoin::{
         consensus,
         hashes::hex::ToHex,
-        secp256k1::Secp256k1,
+        secp256k1::{Secp256k1, Signing, Verification},
+        util::psbt,
         util::sighash::{Prevouts, SighashCache},
-        Address, LockTime, Network, Sequence, Transaction, TxIn, TxOut,
+        Address, LockTime, Network, PackedLockTime, Sequence, Transaction, TxIn, TxOut, Witness,
     },
     coin_select::{coin_select_bnb, CoinSelector, CoinSelectorOpt, WeightedValue},
     descriptor_into_script_iter,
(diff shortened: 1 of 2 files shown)


bdk_core_example/src/main.rs
```diff
@@ -483,3 +465,137 @@ pub fn fully_sync(
...
+            i.witness_utxo
+                .clone()
+                .expect("witness_utxo must be present")
+        })
+        .collect::<Vec<_>>();
+    let prevouts = Prevouts::All(&prevouts);
+
+    // create a short lived transaction
+    let _sighash_tx = psbt.unsigned_tx.clone();
+    let mut sighash_cache = SighashCache::new(&_sighash_tx);
```
LLFourn, line 496: nit: pass this in to each call rather than repeatedly doing it for each input.

LLFourn (changes requested):
Thanks for doing this @afilini. In general the plan module is looking fantastic. I left some questions.  The main improvement I'd love to have is the `DefinteDescriptor` not being necessary to keep around after you've got the plan for it. I'm not sure but I think the only thing you are missing in the data is the merkle root for tap key spends. Then the plan just needs the ability to set the fields on the PSBT with precisely what you need for the plan. 

Do you think this is possible?

I can probably live without it but I still like the idea of the plan being able to do its job without using PSBTs at all but still having full support for PSBTs. e.g. (i) being able to populate a `psbt::Input` with a plan and (ii) being able to extract `Vec<RequiredSig<DefiniteDescriptorKey>>` and `Vec<RequiredPreimage<DefiniteDescriptorKey>>` from a `psbt::Input` that you got from someone else. Then then main signing functionality can be provided without passing in a PSBT but rather just a transaction and its txouts for the inputs and the `RequiredSig`.

PR #73, opened 2022-11-29
Restructure existing crates

rajarshimaitra:
This PR restructures the existing crates around the `bdk_cli` examples.. 

Current example structures looks like this.

 - `bdk_cli_lib`: a library for all the common functions and commands to run a cli app with `bdk_core`.
 - `bdk_esplora_example` : a binary crate that uses `bdk_cli_lib` and  runs a cli app that can sync with an esplora blockchain.


bdk_esplora_example/src/esplora.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        Ok(wallet_scan.update)
+    }
+
+    /// Create a new checkpoint with transactions spending from or to the scriptpubkeys in
+    /// `scripts`.
+    pub fn wallet_scan<K: Ord + Clone, I>(
+        &self,
+        keychains: BTreeMap<K, I>,
+        stop_gap: Option<usize>,
```
evanlinjin, line 47: This is out of the scope of this PR, but I am rather curious about the definition of `stop_gap` here.

In my mind, spks should be dynamically added and stop gap is used to determine when to stop adding spks in which we should scan for.

The implementation here is first creating a large batch of spks, and stop gap is used to determine when to stop scanning.

In my opinion, scanning with dynamically-added-spks would be more useful.

evanlinjin (commented):
I read through the code and it looks good to me. I left some thoughts.

Also, is this PR intended to replace #35?

bdk_esplora_example/src/esplora.rs
rajarshimaitra: Thats a nice idea.. But one caveat could be, IIUC that we don't wanna take in `refs` of the tracker, because of some async issues occurred in bdk prior (as per @LLFourn ).. So we need to have the script source from somewhere to do dynamic feeding.. The electrum should not have to derive scripts..

Maybe some cool loop with feedback system can be designed to dynamically feed scripts into the `scan`, until it hits `stop_gap` amount of `unused` `sync_response`? Something along that line..

But for regular wallet usages with even 1k `used` scripts per keychain, passing on everything in the tracker for sync should not so terrible, in exchange for simpler code.. 

Also note the script iterator size is maxed at the total `derived index` of the tracker.. So a design goal for the tracker should be to be efficient in storing derived indexes, and not have large pool of unused derived scripts..

rajarshimaitra:
> Also, is this PR intended to replace https://github.com/LLFourn/bdk_core_staging/pull/35?

Yes.. Part of it.. #35 also includes `electrum` and `rpc` examples, which will be added as subsequent PRs after this one..

rajarshimaitra:
Removed nit commits.. They are unrelated to this PR, and can be discussed in separate thread..

bdk_esplora_example/src/esplora.rs
LLFourn: > This is out of the scope of this PR, but I am rather curious about the definition of `stop_gap` here.

Same definition as is always used.

> The implementation here is first creating a large batch of spks, and stop gap is used to determine when to stop scanning.

This is not what is happening. The API is an `Iterator`.

> Maybe some cool loop with feedback system can be designed to dynamically feed scripts into the `scan`, until it hits `stop_gap` amount of `unused` `sync_response`? Something along that line..

This cool feedback system is called an `Iterator` which is what we already doing :).

bdk_esplora_example/src/esplora.rs
```diff
@@ -0,0 +1,188 @@
...
+    ) -> Result<ChainGraph<ConfirmationTime>> {
+        let mut dummy_keychains = BTreeMap::new();
+        dummy_keychains.insert((), spks.enumerate().map(|(i, spk)| (i as u32, spk)));
+
+        let wallet_scan = self.wallet_scan(dummy_keychains, None, existing_chain)?;
+
+        Ok(wallet_scan.update)
+    }
+
+    /// Create a new checkpoint with transactions spending from or to the scriptpubkeys in
```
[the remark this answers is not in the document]
LLFourn: This doc comment should be rewritten or removed.

bdk_esplora_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    let change_keychain = if let Some((internal_descriptor, internal_keymap)) = internal {
+        keymap.extend(internal_keymap);
+        keychain_tracker
+            .txout_index
+            .add_keychain(Keychain::Internal, internal_descriptor);
+        Keychain::Internal
+    } else {
+        Keychain::External
+    };
```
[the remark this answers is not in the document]
LLFourn: I think it's probably a good idea to do this one. It could also just be done in the next PR since I guess you'll need the same code for RPC or electrum.

bdk_cli_lib/src/lib.rs
```diff
@@ -0,0 +1,507 @@
...
+/// A structure defining output of a AddressCmd execution.
+#[derive(serde::Serialize, serde::Deserialize)]
+pub struct AddrsOutput {
+    keychain: String,
+    index: u32,
+    addrs: Address,
+    used: bool,
+}
+
+pub fn print_address_details<K, I>(
```
LLFourn, line 158: Most of these things are not simply "printing" things. Better to name `run_address_cmd` or something like that.

LLFourn (approved):
LGTM. Minor nits. Ready to merge when @rajarshimaitra is happy with it.

LLFourn:
Thanks @rajarshimaitra. I've done a big cleanup commit and merged this.

bdk_esplora_example/src/esplora.rs
evanlinjin: @LLFourn good point, sorry I misread the code.

Issue #72, opened 2022-11-25
ChainGraph API for electrum


LLFourn:
1. We want to have the chaingraph enforce that for every txid in `sparse_chain` there is a full tx in the `graph`. To do this we'll have to make those fields private and we'll have to have an error case when the changeset is applied.
2. Electrum needs to fetch incrementally. First it tells you txids you are interested in (associated with your script pubkeys) and then you batch fetch the transactions. We need an API to do this. It makes sense for electrum to return something like a `SparseChain` from the first requests for related txids to the scripts. We then need to turn the txids we're missing into full transactions in a `ChainGraph` that can be applied. Hint: maybe we can have a `missing` function on `ChainGraph` that somehow returns what's missing and then a method on the electrum API to take that set of missing things and turn it into a `ChainGraph` with a batch request.

rajarshimaitra:
I am working on this next..

LLFourn:
Side note this means removing `insert_txid` from chain graph since this is incompatible with this design.

LLFourn:
@evanlinjin I think we should provide this guarantee as the API can result in unintentional unwraps when you think a full tx should be there but it isn't. For electrum I don't think the update process should start with a `ChainGraph` at all:

1. grab txids of given script pubkeys.
2. Figure out which of these we need full transactions for i.e. return something like a `HashSet<Txid>` or maybe a `Vec<(Txid, I)>` where `I` is `TxHeight` in this case.
3. Fetch those full transactions and transform them into `ChainGraph<TxHeight>` with only the transactions in missing set.
4. Apply the `ChainGraph`.

evanlinjin:
@LLFourn what about transactions being confirmed? Because they might already exist in the mempool of the original chain

LLFourn:
@evanlinjin the update `ChainGraph<TxHeight>` will have the updated height of the transactions. I don't see a problem. The `TxHeight` of the transaction in the original chain is ignored as part of this. We only care whether it is present in the original `ChainGraph` or not.

LLFourn:
The best way to go about this might be to do electrum with the current API first. It's not too hard to figure out which transactions are missing and you need to fetch with the current API. Once we can see what that looks like we can try and optimize both the electurm API and the chaingraph API to make it simple to document and explain,

rajarshimaitra:
For ref, I tried the electrum sync with existing API in #77 ..

PR #74, opened 2022-12-02
Unconfirmed consistency

evanlinjin:
WIP


bdk_core/src/sparse_chain.rs
```diff
@@ -211,59 +213,85 @@ impl<I: ChainIndex> SparseChain<I> {
...
+        let invalid_from = self.determine_invalid_from(update)?;
+        let changeset = self.formulate_changeset(update, invalid_from);
+        Result::Ok(changeset)
+    }
+
+    pub fn determine_unconfirming_changeset(
+        &self,
+        graph: &TxGraph,
+        update: &Self,
+        update_graph: &TxGraph,
```
LLFourn, line 274: What's wrong with passing one graph here?
What's wrong with already applying the update in `update` before doing this step?

I think you are trying to make the update and the unconfirmed consistency stuff all happen in one go. I think this might be a mistake from an API perspective. I think you should be able to take any sparse chain and make the unconfirmed txs consistent against a graph.

Maybe we shouldn't even pass in a graph to do this job. What about something like [`Vec::retain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain). e.g `retain_unconfirmed(&mut self, pred: impl Fn(Txid) -> bool) -> ChangeSet`. Then the caller can use a `TxGraph` or whatever they want to figure out whether something conflicts and whether it should be retained.

evanlinjin: > I think you should be able to take any sparse chain and make the unconfirmed txs consistent against a graph.

I don't think this is possible, as we will need the total view of the mempool to "evict" the right transactions.

For example, given a confirmed output `A:0`, which is spent by tx `B`. If `B:0` is an unowned output, and there exists a tx `C` which does CPFP from `B:0`, then `C` would not exist in the sparsechain (as it does not contain or reference a relevant scriptPubKey). Now imagine a RBF attempt to replace `B` (called `B'`).Basically, How can we guarantee consistency if we don't have a total view of all descendant transactions?

evanlinjin: I feel like the best approach is to assume the `update` always contains txs that are either confirmed, or guaranteed to be in the mempool (for that point in time). So anything that conflicts with the update's transactions can be evicted.

bdk_core/src/sparse_chain.rs
```diff
@@ -162,7 +164,9 @@ impl<I: ChainIndex> SparseChain<I> {
         }
     }
 
-    pub fn determine_changeset(&self, update: &Self) -> Result<ChangeSet<I>, UpdateFailure<I>> {
+    /// Determines the height of the original [`SparseChain`] to invalidate from if `update` is to
+    /// be applied.
+    pub fn determine_invalid_from(&self, update: &Self) -> Result<Option<u32>, UpdateFailure<I>> {
```
LLFourn, line 169: Do we need this to be a public API?

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -85,7 +85,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// Scan a single `TxOut` for a matching script pubkey.
     ///
     /// If it matches the index will store and index it.
-    pub fn scan_txout(&mut self, op: OutPoint, txout: &TxOut) {
+    pub fn scan_txout(&mut self, op: OutPoint, txout: &TxOut) -> Option<(K, u32)> {
```
LLFourn, line 88: Is this change a necessary part of this PR?

bdk_core/src/sparse_chain.rs
```diff
@@ -488,13 +494,15 @@ impl<I: ChainIndex> SparseChain<I> {
     serde(crate = "serde_crate")
 )]
 pub struct ChangeSet<I = TxHeight> {
+    pub invalid_from: Option<u32>,
```
LLFourn, line 497: why does changeset have this field?

bdk_core/src/sparse_chain.rs
LLFourn: > Basically, How can we guarantee consistency if we don't have a total view of all descendant transactions?

We're not trying to guarantee it I think but I see what you're getting at. What's the point of doing this if we can't provide guarantees.


> I feel like the best approach is to assume the `update` always contains txs that are either confirmed, or guaranteed to be in the mempool (for that point in time). So anything that conflicts with the update's transactions can be evicted.
> 
> This does not 100% guarantee consistency, but it can guarantee eventual consistency without wrongly evicting transactions.

Great point. Ok I understand your approach much better now.

bdk_core/src/chain_graph.rs
```diff
@@ -63,8 +63,37 @@ impl<I: ChainIndex> ChainGraph<I> {
         &self,
         update: &Self,
     ) -> Result<ChangeSet<I>, sparse_chain::UpdateFailure<I>> {
+        let mut chain_changeset = self.chain.determine_changeset(&update.chain)?;
+
+        // If we are invalidating checkpoints, it signals that a reorg has occured. We modify the
+        // changeset to add back "invalidated" transactions which do not conflict with transactions
+        // introduced by the `update`. We assume that all unconfirmed transactions in the `update`
+        // is included in the mempool.
+        if let Some(invalid_from) = chain_changeset.invalid_from {
```
LLFourn, line 72: It seems that we can easily get `invalid_from` from the checkpoint changes rather than having an explicit field for this.

bdk_core/src/chain_graph.rs
```diff
@@ -63,8 +63,37 @@ impl<I: ChainIndex> ChainGraph<I> {
...
+                .range_txids_by_height(TxHeight::Confirmed(invalid_from)..)
+                .filter_map(|&(_, txid)| self.graph.tx(txid).map(|tx| (txid, tx)))
+                // filter out txs that conflict with update's chain
+                .filter(|&(_, tx)| {
+                    update
+                        .graph
+                        .conflicting_txids(tx)
+                        .find(|(_, txid)| update.chain.tx_index(*txid).is_some())
+                        .is_none()
+                });
```
LLFourn, line 85: It should be simpler to iterate over the new transactions and tell you a list of txids in the current chain that conflict with them. Then attempt to remove them if possible (add the removals to the sparse chain changeset) -- if the new update conflicts with an existing tx (that was not removed in the changeset) then return an error. I don't think we need `invalid_from` at all?

bdk_core/src/chain_graph.rs
```diff
@@ -63,8 +63,37 @@ impl<I: ChainIndex> ChainGraph<I> {
...
+
+            // add back non-conflicting txids to changeset
```
LLFourn, line 87: It would be better if this comment explain *why* we are doing this rather than *what* we are doing.

bdk_core/src/sparse_chain.rs
LLFourn: [edit] I think it can be removed see comment above.

bdk_core/src/chain_graph.rs
```diff
@@ -63,8 +63,37 @@ impl<I: ChainIndex> ChainGraph<I> {
...
+            for (txid, _) in retain_txs {
+                if let Some(tx_index) = chain_changeset.txids.get_mut(&txid) {
+                    tx_index.get_or_insert(I::max_ord_of_height(TxHeight::Unconfirmed));
+                }
+            }
```
LLFourn, line 92: Hmm I suppose this makes sense as a default policy. We probably want to provide ways of doing other policies eventually.

LLFourn (changes requested):
Thanks @evanlinjin. I think the approach is very good and could be improved by doing without `invalid_from`. Your key insight is that we already had the function we needed to do this: `TxGraph::conflicting_txids` and we just needed to apply this when the chaingraph updates.

A comment on style: try to only make comments that explain *why* rather than *what*.

bdk_keychain/src/keychain_txout_index.rs
evanlinjin: No, woops 😅

bdk_core/src/sparse_chain.rs
evanlinjin: No :/

bdk_core/src/chain_graph.rs
```diff
@@ -63,8 +63,26 @@ impl<I: ChainIndex> ChainGraph<I> {
...
         update: &Self,
     ) -> Result<ChangeSet<I>, sparse_chain::UpdateFailure<I>> {
+        let mut chain_changeset = self.chain.determine_changeset(&update.chain)?;
+
+        let conflicting_original_txids = update
+            .chain
+            .iter_txids()
+            // choose original txs that conflicts with the update
+            .flat_map(|&(_, txid)| {
+                let update_tx = update.graph.tx(txid).expect("tx should exist");
```
LLFourn, line 73: I'm not sure we provide the guarantee that this exists? I think that's what #72 was for.

bdk_core/tests/test_sparse_chain.rs
```diff
@@ -291,6 +291,8 @@ fn can_introduce_older_checkpoints() {
 
 #[test]
 fn fix_blockhash_before_agreement_point() {
+    // TODO: This really shouldn't happen. Should we return error if this happens as this may
+    // require to total resync?
```
LLFourn, line 295: I don't think we can return an error because we can't detect it happening? If we can consider making an issue instead of this comment.

LLFourn (approved):
LGTM. A few comments left you may want to check before merging.

PR #77, opened 2022-12-07
Add a electrum syncing example

### LINKED ISSUE #72: ChainGraph API for electrum

rajarshimaitra:
This PR adds an electrum syncing example.. Fixes #72 

The electrum sync is done in two steps.
 - `wallet_txid_scan` returns a candidate `KeychainScan` with TxGraph data.
 - The candidate is compared with tracker's `SparseChain`, and list of new tx data to fetch is derived from the `ChangeSet`.
 - New transactions are fetched in batch and added into the `KeychainScan` update.
 - Apply the final `KeychainScan` on tracker and DB.


rajarshimaitra:
Restructured the commit to handle #72 requirements.

 - Make Chaingraph consistent, by adding few assertions at the `apply_update()` API. 
 - Make the electrum sync in two steps. Get all the txids, find new ones, fetch tx data, apply update..

rajarshimaitra:
Me and @evanlinjin tend to disagree on the approach of the consistency assertion.. While to me it seems the best place to apply it is the update API, Evan thinks there could be a better approach.. Opening this up for discussion and reviews..

rajarshimaitra:
Rebased on current master, and updated the example syncing with recent invariants around `ChainGraph`. Similar to RPC syncing, the first stage returns a `SparseChain`. The nex `to_fetch_tx` id set is derived using `determine_changeset` between the returned sparsechain and existing sparaechain within the tracker..

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,150 @@
...
+    type Error = electrum_client::Error;
+    fn broadcast(&self, tx: &bdk_core::bitcoin::Transaction) -> Result<(), Self::Error> {
+        let _ = self.client.transaction_broadcast(tx)?;
+        Ok(())
+    }
+}
+
+/// We will detect reorg if it has depth less than this
+const REORG_DETECTION_DEPTH: u32 = 100;
+const DEFAULT_STOP_GAP: usize = 10;
```
LLFourn, line 39: There should be no default stop gap.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,150 @@
...
+        &self,
+        scripts: BTreeMap<K, impl Iterator<Item = (u32, Script)>>,
+        stop_gap: Option<usize>,
+        last_known_height: Option<u32>,
+    ) -> Result<(SparseChain, BTreeMap<K, u32>)> {
+        let mut sparse_chain = SparseChain::default();
+
+        // 1. Create checkpoint data from last_known_height - REORG_DEPTH
+        // Reorg situation will he handled at the time of applying this KeychainScan.
+        // If there's reorg deeper than the assumed depth, update process will throw error.
```
LLFourn, line 82: I guess this comment is stale? I don't see an error thrown.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,150 @@
...
+        let current_height = self.get_height()?;
+        let check_from = last_known_height
+            .map(|ht| ht.saturating_sub(REORG_DETECTION_DEPTH))
+            .unwrap_or(0);
```
LLFourn, line 86: If the tip at `last_known_height` has changed then this will always lead to an error in the case the checkpoint before `last_known_height` is more than 100 blocks away. Why not do it the way we did in esplora?

rajarshimaitra: Yes, I was going with the assumption that `SparseChain` will always maintain consecutive sequence of blocks.. 

The rationale behind that was, in sync code we are always adding sequential blocks in checkpoints, and pruning is happening from the start of the chain.. So by that logic we should never have the last common at 100+ blocks away from `last_known_height`? As long as that's true, the other way of throwing error here is more than 100 blocks of reorg, which is unlikely..

But if my assumption is wrong then yes this should not be done in this way, and the esplora way is better. This one is clearly less elegant than the esplora way, but went in this route to reduce passing in any refs from the tracker.. The chain was the only refs required to pass in in the sync function..

bdk_electrum_example/src/electrum.rs
rajarshimaitra: The intention was to say that the error will occur while performing `apply_update` with the `KeychainScan`.. Thats what I meant by the `update process will throw error.`.. Maybe can be made more clear?

bdk_electrum_example/src/electrum.rs
LLFourn: No if there's a default you just don't stop.

bdk_electrum_example/src/electrum.rs
LLFourn: > Yes, I was going with the assumption that `SparseChain` will always maintain consecutive sequence of blocks..

This is definitely not the case. Perhaps it's confusing to call it a "chain". It's meant to be a view of a chain that is sparse not a chain that is sparse.

> But if my assumption is wrong then yes this should not be done in this way, and the esplora way is better. This one is clearly less elegant than the esplora way, but went in this route to reduce passing in any refs from the tracker.. The chain was the only refs required to pass in in the sync function..

You can just clone the checkpoints to avoid taking a ref. There might be other ways to design this process but the way it works is to limit the number of checkpoints in the sparse chain with `checkpoint_limit` so they can be quickly cloned and passed in here without taking a ref. This seems to be ideal since the your checkpoint limit will be a function of your confidence of the persistence of the underlying blockchain and the memory available on your platform -- these are not things that can change at runtime and should be fixed in the config.

bdk_electrum_example/src/electrum.rs
LLFourn: Ah sorry I didn't read this. Let's design it so this error case doesn't exist :). I don't see how you're meant to recover from the error with this API.

bdk_electrum_example/src/electrum.rs
rajarshimaitra: Okay.. Ya that makes sense to me.. So I will revert it back to the esplora way..

evanlinjin (changes requested):
Good work so far!

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,150 @@
...
+                                TxHeight::Confirmed(history_result.height as u32),
+                            );
+                        } else {
+                            return (history_result.tx_hash, TxHeight::Unconfirmed);
+                        };
+                    })
+                    .collect::<Vec<(Txid, TxHeight)>>();
+
+                if history.is_empty() {
+                    unused_script_count += 1;
```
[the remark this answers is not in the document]
rajarshimaitra: At the start of the loop for every keychain..

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,169 @@
...
+                .zip(txid_changeset.into_iter().map(|(_, index)| index))
+            {
+                scan_update.update.insert_tx(tx, index)?;
+            }
+
+            // Apply the full scan update
+            let changeset = tracker.determine_changeset(&scan_update)?;
+            db.append_changeset(&changeset)?;
+            tracker.apply_changeset(changeset);
+            debug!("sync completed!!")
```
LLFourn, line 156: Let's put the duplicated code between sync and scan into a function.

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,169 @@
...
+                let _ = scan_update.update.insert_checkpoint(BlockId {
+                    height: *height,
+                    hash: *hash,
+                })?;
+            }
+
+            // New Txids that we don't know the full data of.
+            let txid_changeset = tracker
+                .chain()
+                .determine_changeset(&new_sparsechain)?
```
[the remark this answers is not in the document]
rajarshimaitra: I don't follow.. Its a `SparseChain::ChangeSet`, that only gives the `txid` changes. So we don't have tx_additions at this point. 

But I see your point, and added another filter to not fetch redundant transactions in the sync. But this causes another complexity, as in theory we can have partial transactions in the graph, and `insert_tx` only allows full transaction data.

I have added the new logic in latest push, and marked the above as TODO.

bdk_electrum_example/src/main.rs
rajarshimaitra: Good idea.. Was thinking exactly that..

rajarshimaitra:
Updated with the above suggestions.

 - No default reorg depth mechanism. Takes in a ref of existing chain, and inserts checkpoint until the last common check point is found. And then push the current tip into update. In case of reorg, all checkpoints after the last common block is added into the update structure, and while `SparseChain::apply_update` new checkpoints will replace old checkpoints. No error is thrown.
 
 - No default stop gap.
 
 - Transaction fetching is done opportunistically, for only txids that we haven't seen before. For transactions that only changed in `Index` tx data is fetched from existing `TxGraph` of the tracker. 
 
 Q: What to do when we only have `TxNode::Partial` in graph, and index changes of that transaction? 2 possible options. 
  - We can fetch such transactions again from the blockchain before applying update.
  - Have mechanism to `insert_tx()` with `TXNode::Partial`
  
- A new function added in the main to collect all the common code between `Scan` and `Sync`.

LLFourn:
> Q: What to do when we only have `TxNode::Partial` in graph, and index changes of that transaction? 2 possible options.
> 
>     * We can fetch such transactions again from the blockchain before applying update.
> 
>     * Have mechanism to `insert_tx()` with `TXNode::Partial`

Haven't reviewed just comment: we don't keep the index for `TxNode::Partial`.

rajarshimaitra:
> Haven't reviewed just comment: we don't keep the index for TxNode::Partial.

Done..

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,151 @@
...
+                    unused_script_count = 0;
+                }
+
+                if unused_script_count >= stop_gap.unwrap_or(usize::MAX) {
+                    break;
+                }
+            }
+
+            for (txid, index) in script_history_txid {
+                sparse_chain.insert_tx(txid, index)?;
```
LLFourn, line 143: Deal with this error here.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,151 @@
...
+            let mut script_history_txid = Vec::<(Txid, TxHeight)>::new();
+
+            loop {
+                let (index, script) = scripts.next().expect("its an infinite iterator");
+
+                let history = self
+                    .script_get_history(&script)?
+                    .iter()
+                    .map(|history_result| {
+                        if history_result.height > 0 {
```
LLFourn, line 118:
```suggestion
                        if history_result.height > 0 && history_result.height <= tip_height {
```

we don't want to add as confirmed things above the tip height. Fine to add them to unconfirmed. This just means a new block has been confirmed since you started. I'm not sure what we did in esplora but we should try and do the same thing there too.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,151 @@
...
+                }
+            }
+
+            for (txid, index) in script_history_txid {
+                sparse_chain.insert_tx(txid, index)?;
+            }
+
+            keychain_index_update.insert(keychain, last_active_index);
+        }
+
```
LLFourn, line 148: Must check with electrum that the chain at the end still contains `tip_hash`. If not return an error indicating there has been a re-org during scan.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,151 @@
...
+            let current_hash = self
+                .client
+                .block_header(existing_height as usize)?
+                .block_hash();
+            sparse_chain
+                .insert_checkpoint(BlockId {
+                    height: existing_height,
+                    hash: current_hash,
+                })
+                .expect("should not collide");
```
LLFourn, line 86: This is a very interesting "should". When does this panic?

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+            // Wallet scan returns a sparse chain that contains new All the BlockIds and Txids
+            // relevant to the wallet, along with keychain index update if required.
+            let (new_sparsechain, keychain_index_update) =
+                client.wallet_txid_scan(scripts, Some(stop_gap), tracker.chain().checkpoints())?;
+
+            keychain_scan.last_active_indexes = keychain_index_update;
+
+            // Inserting everything from the new_sparsechain should be okay as duplicate
+            // data would be rejected at the time of update application.
+            for (height, hash) in new_sparsechain.checkpoints() {
```
LLFourn, line 115: Why doesn't this just return `BlockId`s :thinking: . I swear it used to.

bdk_electrum_example/src/main.rs
LLFourn: It seems like there is still a bit of duplicate code. Isn't it the case that everything after getting the sparse chain and applying the last active indexes will be the same?

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+        .chain(existing_txs)
+        .collect::<Vec<_>>();
+
+    Ok(transaction_update)
+}
+
+fn main() -> anyhow::Result<()> {
+    let (args, keymap, mut tracker, mut db) = bdk_cli::init::<ElectrumCommands, _>()?;
+
+    let client = ElectrumClient::new("ssl://electrum.blockstream.info:60002")?;
```
LLFourn, line 84: needs to be different for different networks?

LLFourn (changes requested):
Thanks @rajarshimaitra. Looks great. A few minor requests to finish this off. The main thing that's missing is to protect against re-org during sync.

bdk_electrum_example/src/electrum.rs
rajarshimaitra: The in sync reorg situation is handled as described above.. https://github.com/LLFourn/bdk_core_staging/pull/77#discussion_r1049572798

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        // Insert the new tip
+        let (tip_height, tip_hash) = self.get_tip()?;
+        sparse_chain
+            .insert_checkpoint(BlockId {
+                height: tip_height,
+                hash: tip_hash,
+            })
+            .expect("Should not collide");
```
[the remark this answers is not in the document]
rajarshimaitra: Same as above..

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// Scan for a keychain tracker, and create an initial [`KeychainScan`] candidate update.
+    /// This update will only contain [`Txid`]s in SparseChain, and no actual transaction data.
+    ///
+    /// User needs to fetch the required transaction data and update the [`KeychainScan`] before applying it.
+    pub fn wallet_txid_scan<K: Ord + Clone>(
+        &self,
+        scripts: BTreeMap<K, impl Iterator<Item = (u32, Script)>>,
+        stop_gap: Option<usize>,
+        local_chain: &BTreeMap<u32, BlockHash>,
+    ) -> Result<(SparseChain, BTreeMap<K, u32>)> {
```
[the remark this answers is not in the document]
rajarshimaitra: Added custom result.. Does it make sense to use `thiserror` for this case? Would reduce boiler plates.. If yes, will do that on a separate PR.

bdk_electrum_example/src/main.rs
rajarshimaitra: I feel stupid but I can't seem to find any docs out there for electrum with other network ports.. Any help??

rajarshimaitra:
> It seems like there is still a bit of duplicate code. Isn't it the case that everything after getting the sparse chain and applying the last active indexes will be the same?

Yes.. I tried to add this into a separate function but that's making the code more complicated, because many trait bounds needs to be satisfied for the `KeychainStore::append_changeset()`, which are handled in `bdk_cli`.. It still could be done, but after doing it it made the code more complex than it needs to be, and thought would be detriment for readers reading this as example, and its only removing 3 lines. So kept it as it is for now..

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,236 @@
...
+        let mut reorgred_tx = Vec::new();
+
+        // Fetch Keychain's last_active_index and all related txids.
+        // Add them into the KeyChainScan
+        for (keychain, mut scripts) in scripts.into_iter() {
+            let mut last_active_index = 0;
+            let mut unused_script_count = 0usize;
+            let mut script_history_txid = Vec::<(Txid, TxHeight)>::new();
+
+            loop {
```
LLFourn, line 158: This should be a `while let Some((index, script)) = scripts.next()`

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,236 @@
...
+                if let Some((index, script)) = scripts.next() {
+                    let history = self
+                        .script_get_history(&script)?
+                        .iter()
+                        .map(|history_result| {
+                            if history_result.height > 0
+                                && (history_result.height as u32) < tip_height
```
LLFourn, line 165: `(history_result.height as u32) <= tip_height`

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,236 @@
...
+    BlockId, TxHeight,
+};
+use electrum_client::{Client, Config, ElectrumApi};
+
+#[derive(Debug)]
+pub enum ElectrumError {
+    Client(electrum_client::Error),
+    InsertTx(InsertTxErr),
+    Update(UpdateFailure),
+    InsertCheckpoint(InsertCheckpointErr),
```
LLFourn, line 16: Each of these errors needs docs. It should be clear how a user is meant to recover from each of them. If I get a client error I guess I know that something went wrong connecting to the electrum client but with the others I have no clue what it means.

bdk_electrum_example/src/electrum.rs
LLFourn: You are throwing the error on `insert_checkpoint` so doesn't this just return with an error if there was a re-org (since `insert_checkpoint` returns an error if the hashes don't match). It seems simpler just to check that the hash at the start tip height is the same and throw an error if not.

It looks like you are trying to do your best even in the face of a re-org and still return a sparse chain. This is a cool idea. It might require some more thought though. For now I suggest just returning a `ReOrg` error if there was a re-org while you are doing the scan and we can make an issue about how to do better than this.

Inserting some extra checkpoints is ok -- we do that in esplora but in that case we have a single call to fetch the most recent ones. It might not be worth it here.

bdk_electrum_example/src/electrum.rs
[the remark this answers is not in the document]
LLFourn: I don't really mind whether we use thiserror in the blockchain example modules.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,236 @@
...
+                                return (history_result.tx_hash, TxHeight::Unconfirmed);
+                            };
+                        })
+                        .collect::<Vec<(Txid, TxHeight)>>();
+
+                    if history.is_empty() {
+                        unused_script_count += 1;
+                    } else {
+                        last_active_index = index;
+                        script_history_txid.extend(history.iter());
```
LLFourn, line 181: why not just insert into the sparse chain right here.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,236 @@
...
+        // Add them into the KeyChainScan
+        for (keychain, mut scripts) in scripts.into_iter() {
+            let mut last_active_index = 0;
+            let mut unused_script_count = 0usize;
+            let mut script_history_txid = Vec::<(Txid, TxHeight)>::new();
+
+            loop {
+                if let Some((index, script)) = scripts.next() {
+                    let history = self
+                        .script_get_history(&script)?
```
LLFourn, line 161: shouldn't we be using `batch_script_get_history` for efficiency here?

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,236 @@
...
+                } else {
+                    break;
+                }
+            }
+
+            for (txid, index) in script_history_txid {
+                if let Err(err) = sparse_chain.insert_tx(txid, index) {
+                    match err {
+                        InsertTxErr::TxTooHigh => {
+                            /* Don't care about new transactions confirmed while syncing */
```
LLFourn, line 197: since you checked that no tx is higher than tip I think this should be `unreachable!(..)`.

bdk_electrum_example/src/electrum.rs
rajarshimaitra: We can, but currently we are getting all the scripts using `iter_all_script_pubkeys_by_keychain()`, which calls  `descriptor_into_script_iter()` which is an iterator over all the scripts that can be derived from a descriptor. So if we call `batch_script_history`, it will work for the `Sync` call, but for `Scan` it will try to fetch for `2^32` values for wild card descriptors which can be a pretty massive call..

bdk_electrum_example/src/electrum.rs
rajarshimaitra: Done..

bdk_electrum_example/src/electrum.rs
rajarshimaitra: Good point.. Done..

bdk_electrum_example/src/electrum.rs
rajarshimaitra: Makes sense.. Done..

rajarshimaitra:
> Each of these errors needs docs. It should be clear how a user is meant to recover from each of them. If I get a client error I guess I know that something went wrong connecting to the electrum client but with the others I have no clue what it means.

On further thinking these error variants should not be there. As users don't know what to do in those cases. I think in the previous code we could have just `expect` the result because those were emitted from an empty sparsechain.

> It looks like you are trying to do your best even in the face of a re-org and still return a sparse chain. This is a cool idea. It might require some more thought though. For now I suggest just returning a ReOrg error if there was a re-org while you are doing the scan and we can make an issue about how to do better than this.

Yes that was my initial intention. As users won't know what to do in case of a in-sync reorg. And I don't think there's any recourse available for user in the current API. So ideally  `wallet_txid_scan` should only throw client error, and everything else should be handled by the logic itself. In case of reorg we should be able to still return a sparsechain..

But I went ahead with your suggestion and removed that extra logic, and now it just throws a reorg error in case reorg is detected. Added a TODO to note this point for further works.

bdk_electrum_example/src/electrum.rs
rajarshimaitra: Done..

bdk_electrum_example/src/electrum.rs
rajarshimaitra: Removed these error variants in latest push..

bdk_electrum_example/src/electrum.rs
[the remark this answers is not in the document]
rajarshimaitra: Hmm.. I think so ya.. Returning a Reorg error in that case??

bdk_electrum_example/src/main.rs
LLFourn: I think you can at least just return the `new_sparsechain` from the match arms and do the rest outside of the match statement.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        // Insert the new tip
+        let (tip_height, tip_hash) = self.get_tip()?;
+        if sparse_chain
+            .insert_checkpoint(BlockId {
+                height: tip_height,
+                hash: tip_hash,
+            }).is_err() {
+                // This means our existing chain tip has been reorged out.
```
LLFourn, line 123: It means the tip you just added to has changed since you last requested it at the end of the for loop above. You use "existing" to refer to the one in `local_chain` above but that's not what's happening here. Please clear up this comment.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,205 @@
...
+                } else {
+                    last_active_index = index;
+                    unused_script_count = 0;
+
+                    for (txid, index) in history {
+                        if let Err(err) = sparse_chain.insert_tx(txid, index) {
+                            match err {
+                                InsertTxErr::TxTooHigh => {
+                                    /* We should not encounter this error as we ensured TxHeight <= tip_height */
+                                    unreachable!();
```
LLFourn, line 164: Put the comment above in the unreachable statement as a string.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        // Insert the new checkpoints
+        for (hash, height) in reorged_headers {
+            if sparse_chain
+                .insert_checkpoint(BlockId { height, hash })
+                .is_err()
+            {
+                return Err(ElectrumError::Reorg(height));
+            };
+        }
```
LLFourn, line 201: Simplify this. I think we should just check that our tip hasn't changed here. If it has return the re-org error. That's enough for now.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,205 @@
...
+    bitcoin::{BlockHash, Script, Txid},
+    sparse_chain::{InsertTxErr, SparseChain},
+    BlockId, TxHeight,
+};
+use electrum_client::{Client, Config, ElectrumApi};
+
+#[derive(Debug)]
+pub enum ElectrumError {
+    Client(electrum_client::Error),
+    Reorg(u32),
```
LLFourn, line 14: I don't think this height thing is useful. What is the caller meant to do with it?

Some doc comments here would be good to explain how to recover from the two different error kinds. `Reorg` should just be call the function again.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        // Fetch Keychain's last_active_index and all related txids.
+        // Add them into the KeyChainScan
+        for (keychain, mut scripts) in scripts {
+            let mut last_active_index = 0;
+            let mut unused_script_count = 0usize;
+
+            while let Some((index, script)) = scripts.next() {
+                let history = self
+                    .script_get_history(&script)?
```
LLFourn, line 138: Lets use `batch_script_get_history` here.

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+                let _ = scan_update.update.insert_checkpoint(BlockId {
+                    height: *height,
+                    hash: *hash,
+                })?;
+            }
+
+            // Fetch the new and old transactions to be added in update structure
+            for (tx, index) in fetch_transactions(&new_sparsechain, &client, &tracker)? {
+                scan_update.update.insert_tx(tx, index)?;
+            }
```
LLFourn, line 203: Can you try simplifying the above with the new APIs we added in https://github.com/LLFourn/bdk_core_staging/pull/92?

You can use `SpaseChain::changeset_additions` to figure out what the changeset is between your local sparse chain and the new sparse chain.. Fetch those full transactions and then call `ChainGraph::inflate_changeset` to turn your `sparse_chain::ChangeSet` into a `ChainGraph::ChangeSet`. You can then set the `KeychainChangeSet`'s `chain_graph` field with this value and just apply that changeset.

LLFourn (commented):
I just took another pass at this. The list of things I've proposed to fix here is final now I think.

bdk_electrum_example/src/electrum.rs
rajarshimaitra: I didn't do this one because of the comment here https://github.com/LLFourn/bdk_core_staging/pull/77#discussion_r1050440828 and also couldn't figure how to use `stop_gap` in that case. Let me know if you intend to fetch for 2^32 scripts anyway and should we remove stop gap?? 

The other way is by calling `script_pubkeys_by_keychain` which will give us iterator over only derived scripts.. Let me know if that's the way we wanna go here..

rajarshimaitra:
Thanks @LLFourn for the comments.. Updated as suggested and used the new API.. modulo https://github.com/LLFourn/bdk_core_staging/pull/77#discussion_r1054936229..  Code looks much simpler..

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+
+fn complete_update<K: Debug + Clone + Ord, I: ChainIndex>(
+    new_sparsechain: &SparseChain<I>,
+    tracker: &KeychainTracker<K, I>,
+    client: &ElectrumClient,
+) -> Result<KeychainChangeSet<K, I>> {
+    let sparsechain_changeset = match tracker.chain().determine_changeset(new_sparsechain) {
+        Ok(cs) => cs,
+        Err(update_failure) => {
+            return Err(anyhow!(
```
LLFourn, line 50: just use `Anyhow::Context::context` to add these annotations rather than matching them.

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,197 @@
...
+            keychain_index_update.insert(keychain, last_active_index);
+        }
+
+        // Check for Reorg during the above sync process
+        let tip_blockid = {
+            let (height, hash) = self.get_tip()?;
+            BlockId { height, hash }
+        };
+
+        if sparse_chain.latest_checkpoint().expect("must exist") != tip_blockid {
```
LLFourn, line 191: This tip changing doesn't mean there has been a re-org. The block hash at the same height as `latest_checkpoint` changing means there has been a re-org.

bdk_electrum_example/src/electrum.rs
LLFourn: Sorry I missed that comment. Look at what we do in electrum impl in bdk already. You just take a bunch of scripts up to `batch_size` which the caller can pass in and batch it in that way.

bdk_electrum_example/src/electrum.rs
LLFourn: Sorry I responded in the linked convo.

bdk_electrum_example/src/main.rs
LLFourn: Thanks this looks great.

LLFourn (commented):
Look great @rajarshimaitra. Main things:

- use batch calls to electrum
-  Intead of matching every error and then throwing a new anyhow error just use `anyhow::Context::context`
- Remove function `complete_update` in favor of just getting the `new_sparsechain` from the match arms and doing the update from there like what @evanlinjin does here: https://github.com/LLFourn/bdk_core_staging/pull/107/files#diff-6f57072cac1464050942f5bf972a9a8f76cb82a151f93f5b5d16cab1515f925aR74

bdk_electrum_example/src/main.rs
rajarshimaitra: The major reason I couldn't apply `context` or just pass on the error is the error type returned here isn't `std::err:Error`. So compiler is complaining on trait bound.. That's the reason for such awkward error handling.. Same with other places where the return type isn't `std::Error`..  

Here's the full error
```
error[E0599]: the method `context` exists for enum `Result<(), (KeychainChangeSet<Keychain, _>, HashSet<bdk_core::bitcoin::Txid>)>`, but its trait bounds were not satisfied
   --> bdk_electrum_example/src/main.rs:116:57
    |
116 |             tracker.apply_changeset(keychain_changeset).context("Changeset Application to tracker failed")?;
    |                                                         ^^^^^^^ method cannot be called on `Result<(), (KeychainChangeSet<Keychain, _>, HashSet<bdk_core::bitcoin::Txid>)>` due to unsatisfied trait bounds
    |
   ::: /home/raj/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:504:1
    |
504 | pub enum Result<T, E> {
    | --------------------- doesn't satisfy `_: bdk_cli::anyhow::Context<(), (KeychainChangeSet<Keychain, _>, HashSet<bdk_core::bitcoin::Txid>)>`
    |
    = note: the following trait bounds were not satisfied:
            `(KeychainChangeSet<Keychain, _>, HashSet<bdk_core::bitcoin::Txid>): anyhow::context::ext::StdError`
            which is required by `Result<(), (KeychainChangeSet<Keychain, _>, HashSet<bdk_core::bitcoin::Txid>)>: bdk_cli::anyhow::Context<(), (KeychainChangeSet<Keychain, _>, HashSet<bdk_core::bitcoin::Txid>)>`
```
The work around for that is either this awkward handling, or create dedicated error type to return those errors. I think we should do the later, but would prefer to do them in a separate PR..

bdk_electrum_example/src/electrum.rs
rajarshimaitra: Yes makes sense.. Updated with hash matching of latest checkpoint that we have..

rajarshimaitra:
@LLFourn updated with the suggestions, module https://github.com/LLFourn/bdk_core_staging/pull/77#discussion_r1056064131. 

Let me know if this looks good and you want the the commit history squashed..

rajarshimaitra:
Sqaushed and rebased on master..

LLFourn:
I just did some work to get this working:

- break at the right time in the loop for stop gap
- add signet electrum server
- print out scnning progress
- Make client configurable
- Stop returning the sparse chain from inflate_changeset because it
  makes it harder to ? the error and you can just clone it if you still
  need it on failure.

Check it out @rajarshimaitra.

This is working and probably ready to merge with final review from @evanlinjin.

LLFourn (approved):

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,212 @@
...
+                        .map(|history_result| {
+                            if history_result.height > 0
+                                && (history_result.height as u32) <= tip_height
+                            {
+                                (
+                                    history_result.tx_hash,
+                                    TxHeight::Confirmed(history_result.height as u32),
+                                )
+                            } else {
+                                (history_result.tx_hash, TxHeight::Unconfirmed)
```
evanlinjin, line 170: Smart.

rajarshimaitra:
ACK https://github.com/LLFourn/bdk_core_staging/commit/35319072cdfc2de982a4ccb69714c6aa1a601203

Thanks, @LLFourn for the fixes.. Code looks much better.. 

Added a nit fix.. 

Testing out the full thing..

evanlinjin:
@rajarshimaitra You forced-pushed away the changes...

rajarshimaitra:
Holy shit..

evanlinjin:
@rajarshimaitra Fixed.

rajarshimaitra:
Sorry dude.. Intended to just push a nit.. Fat fingered..

bdk_electrum_example/src/electrum.rs
```diff
@@ -0,0 +1,212 @@
...
+                            }
+                        })
+                        .collect::<Vec<(Txid, TxHeight)>>();
+
+                    if txid_list.is_empty() {
+                        unused_script_count += 1;
+                    } else {
+                        last_active_index = index;
```
evanlinjin, line 178: Given the API, it is not guaranteed that the scripts will be given in order by index (`impl Iterator<Item = (u32, Script)>`). Should we check if `index` is greater than `last_active_index` before changing?

bdk_electrum_example/src/main.rs
```diff
@@ -0,0 +1,191 @@
...
+        scan_option: ScanOption,
+    },
+}
+
+#[derive(Parser, Debug, Clone, PartialEq)]
+pub struct ScanOption {
+    /// Set batch size for each script_history call to electrum client
+    #[clap(long, default_value = "25")]
+    pub batch_size: usize,
+}
```
evanlinjin, line 44: NIT: If we can add an optional electrum_url flag, it would be helpful imo.

evanlinjin (approved):
ACK 35319072cdfc2de982a4ccb69714c6aa1a601203

Just some nits

bdk_electrum_example/src/main.rs
evanlinjin: I can't find a way to do this elegantly so I will skip this.

PR #78, opened 2022-12-07
Invalidation consistency tests

### LINKED ISSUE #76: Tests for mempool consistency
We merged #74 but didn't really add any tests to properly enforce the new behaviors. 

The main thing I can see is missing is testing that a transaction that was in a stale block which conflicts with an update transaction does not make it into the mempool.

I think #74 also introduced a bug where an update can evict a transaction in the chain without invalidating its block by having a new transaction that conflicts with it. We shouldn't allow this and this should mean the update is invalid. Test for this too.

evanlinjin:
Closes #76 

* Ensure "invalidated" txids will move to `TxHeight::Unconfirmed`.
* Ensure `ChangeSet` only records changes that actually occur.

### DIFF
--- a/bdk_core/src/chain_graph.rs
+++ b/bdk_core/src/chain_graph.rs
@@ -63,22 +63,33 @@ impl<I: ChainIndex> ChainGraph<I> {
         &self,
         update: &Self,
     ) -> Result<ChangeSet<I>, sparse_chain::UpdateFailure<I>> {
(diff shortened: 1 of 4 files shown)


evanlinjin:
@LLFourn sorry I will merge this without your review because what @rajarshimaitra and myself is working on right now relies on this.

PR #79, opened 2022-12-07
Add a RPC syncing example.

rajarshimaitra:
This PR adds a RPC syncing example with "create a core wallet, and call `listtransaction`" approach.   

Similar to electrum, RPC sync is done in two steps.

 - get all the newly added txid from and initial wallet sync. This stage also includes the reorg detection.

 - Fetch the txs not already stpred in the TxGraph and complete the final update structure.

 - Apply the update to both tracker and database.

 Note: The current approach of RPC syncing with a corresponding internal
 core wallet, cannot do random script pubkey scanning.


rajarshimaitra:
@evanlinjin, updated the example to work with the latest invariant around `ChainGraph`. The `wallet_scan` returns `SparseChain` which is then used to determine the newly found txids.. This allows us to keep the two stage syncing process working, without violating the `ChainGraph` restrictions..

rajarshimaitra:
Latest push

 - Rebased on master. Fixed conflicts.
 - Modified the RPC syncing logic with the similar internal APIs used for electrum and esplora.

@evanlinjin @LLFourn This is now ready for a look..

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,349 @@
...
+                break;
+            }
+            page += 1;
+        }
+
+        for (index, txid) in txids_to_update {
+            if let Err(failure) = sparse_chain.insert_tx(txid, index) {
+                match failure {
+                    sparse_chain::InsertTxError::TxTooHigh { .. } => {
+                        unreachable!("We should not encounter this error as we ensured tx_height <= tip.height");
```
evanlinjin, line 304: I don't think this check is implemented yet.

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,349 @@
...
+                    }
+                    sparse_chain::InsertTxError::TxMovedUnexpectedly { .. } => {
+                        /* This means there is a reorg, we will handle this situation below */
```
evanlinjin, line 307: It seems the situation is not handled. Maybe it is better to break here?

bdk_rpc_wallet_example/src/rpc.rs
rajarshimaitra: Thanks for the observation. I missed it. Added the logic in the `list_tx_result` calculation above.

bdk_rpc_wallet_example/src/rpc.rs
rajarshimaitra: I think the break happens in the next segment `// Check for Reorg during the above sync process`. I do think we should do better than just error in such cases. But this is how it's done for other examples, so I followed the same here too. 

A few lines above, when we first check for `Reorg during sync`, we call the scan function recursively. Maybe we can do the same for the case after extracting the list of transactions? i.e, if there's again a reorg during sync, call scan again? 

Open to suggestions here.

rajarshimaitra: Reorg during sync is a very low probability situation, and when the user gets this error, they should just run the scan again. In that perspective, it makes sense to call the scan again internally rather than throwing the error?

rajarshimaitra:
closing in favour of #170

PR #80, opened 2022-12-09
Enforce invariants for `ChainGraph` structure

evanlinjin:
The `ChainGraph` structure now ensures there are no "floating txids". All txids mentioned in the internal chain must exist in the internal graph.

### DIFF
--- a/bdk_core/src/chain_graph.rs
+++ b/bdk_core/src/chain_graph.rs
@@ -19,8 +19,8 @@ use core::fmt::Debug;
 /// from the graph.
 #[derive(Clone, Debug, PartialEq)]
 pub struct ChainGraph<I = TxHeight> {
-    pub chain: SparseChain<I>,
-    pub graph: TxGraph,
+    chain: SparseChain<I>,
+    graph: TxGraph,
 }
 
 impl<I> Default for ChainGraph<I> {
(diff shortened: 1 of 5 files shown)


bdk_core/src/chain_graph.rs
```diff
@@ -32,25 +32,38 @@ impl<I> Default for ChainGraph<I> {
...
+        let chain_changed = match index {
+            Some(index) => self.chain.insert_tx(tx.txid(), index)?,
+            None => false,
+        };
+        let graph_changed = self.graph.insert_tx(tx);
+
+        Ok(InsertOk {
+            chain: chain_changed,
+            graph: graph_changed,
+        })
```
LLFourn, line 60: I can see why you are returning this but I don't think we need it. You can figure out whether a tx is changing height from the outside in the odd case you really need to know this. Changes where you need to record them are done with changesets.

bdk_core/src/chain_graph.rs
```diff
@@ -32,25 +32,38 @@ impl<I> Default for ChainGraph<I> {
...
 impl<I: ChainIndex> ChainGraph<I> {
     pub fn insert_tx(
         &mut self,
         tx: Transaction,
-        index: I,
-    ) -> Result<bool, sparse_chain::InsertTxErr> {
-        let changed = self.chain.insert_tx(tx.txid(), index)?;
-        self.graph.insert_tx(tx);
-        Ok(changed)
+        index: Option<I>,
```
LLFourn, line 49: Hmm interesting choice making this an `Option`. I think it's a fine choice. Is there a situation you had in mind that you would want to add a transaction without knowing its height.

bdk_core/src/chain_graph.rs
```diff
@@ -19,8 +19,8 @@ use core::fmt::Debug;
 /// from the graph.
 #[derive(Clone, Debug, PartialEq)]
 pub struct ChainGraph<I = TxHeight> {
-    pub chain: SparseChain<I>,
-    pub graph: TxGraph,
+    chain: SparseChain<I>,
```
LLFourn, line 22: chain has a checkpoint limit that you need to be able to set. Either make setting it possible on the chain graph or make an issue about discussing this if you think it needs more thought.

Given all our changes are checkpoint limits still needed?

LLFourn (approved):
Nice work. I think this is fine as is. I left some comments.

bdk_core/src/chain_graph.rs
evanlinjin: Yes, I think it will be cool to explicitly add evicted transactions. I.e. helps for listing transactions that a certain one has replaced.

bdk_core/src/chain_graph.rs
evanlinjin: I'm not sure, I'll expose the method to set checkpoint limits.

bdk_core/src/chain_graph.rs
evanlinjin: I think I will leave this as is for now

PR #81, opened 2022-12-09
Electrum sync 2

rajarshimaitra:
This is an alternate version of the electrum syncing example on top of #80 .. The update chain graph now contains all the wallet related transaction data, both new and old.. And sync isn't performed in two steps because of the invariant, and the result I think is very similar to esplora syncing..


rajarshimaitra:
Closing this in favor of #77 .. This dumb way of syncing doesn't make sense when we can do step by step syncing using sparsechain.

PR #82, opened 2022-12-12
Remove invalid_from in return value of sparse chain update

LLFourn:
#78 introduced an invalid_from tuple return value to `SparseChain::determine_changeset`. I don't see why you need it so I removed it. I passed the test for it without it so maybe it's safe to remove :)

I'd prefer to merge this before the electrum stuff because it actually directly uses `SparseChain::dertermine_changeset`.


evanlinjin (commented):
We need to allow eviction of originally-confirmed transactions that are invalidated by update.

Check test added in: fad5a0a8a6445f3514e399a5f33a1abb72a18423

I suggest that we go with the new logic introduced in #86

PR #86, opened 2022-12-13
Sparse chain remove invalid from return 2

evanlinjin:
This replaces the work, and closes #82

### DIFF
--- a/bdk_core/src/chain_graph.rs
+++ b/bdk_core/src/chain_graph.rs
@@ -76,38 +76,63 @@ impl<I: ChainIndex> ChainGraph<I> {
     }
 
     /// Calculates the difference between self and `update` in the form of a [`ChangeSet`].
-    pub fn determine_changeset(
-        &self,
-        update: &Self,
-    ) -> Result<ChangeSet<I>, sparse_chain::UpdateFailure<I>> {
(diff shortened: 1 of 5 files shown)


bdk_core/src/chain_graph.rs
```diff
@@ -82,46 +82,53 @@ impl<I: ChainIndex> ChainGraph<I> {
...
+            .flat_map(|(update_ind, update_txid)| {
                 let update_tx = update.graph.tx(*update_txid).expect("must exist");
                 self.graph
                     .conflicting_txids(update_tx)
-                    .filter_map(|(_, existing_txid)| {
-                        self.chain
-                            .tx_index(existing_txid)
-                            .map(|existing_index| (*update_txid, (existing_index, existing_txid)))
+                    .filter_map(move |(_, original_txid)| {
+                        self.chain.tx_index(original_txid).map(|original_ind| {
```
LLFourn, line 96: we're using "existing" and "original" to mean the same thing here. Can we choose one. Or perhaps `conflicting_tx` would be better.

LLFourn (approved):
Nice work spotting the bug! Merge when ready.

bdk_core/src/chain_graph.rs
evanlinjin: Updated.

PR #87, opened 2022-12-13
Allow custom subcommands for `bdk_cli_lib::Args`

evanlinjin:
The nature of the sync logic between electrum-based APIs and block-by-block APIs (such as bitcoin-core-chain and compact-block-filters) are quite different, and hence, subcommands will vary.


evanlinjin:
Replaced by #88

PR #88, opened 2022-12-13
Let example be in control of its commands

LLFourn:
Alternative to https://github.com/LLFourn/bdk_core_staging/pull/87 which uses clap features to properly separate things. The example itself is now in control of what its chain specifc commands are.

I think the new argument added in #87 should instead be a argument to whatever subcommand is going to be added in the rpc example.

Let's merge this and update the examples before proceeding.


evanlinjin (approved):
This is much better than #87

PR #89, opened 2022-12-14
RPC chain example

evanlinjin:
RPC Chain example.

### DIFF
--- a/bdk_keychain/src/keychain_txout_index.rs
+++ b/bdk_keychain/src/keychain_txout_index.rs
@@ -1,3 +1,4 @@
+use alloc::vec::Vec;
 use bdk_core::{
     bitcoin::{secp256k1::Secp256k1, OutPoint, Script, TxOut},
     collections::*,
@@ -79,14 +80,14 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     ///
     /// [`ForEachTxout`]: bdk_core::ForEachTxout
     pub fn scan(&mut self, txouts: &impl ForEachTxout) {
-        self.inner.scan(txouts);
+        self.inner.scan(txouts)
     }
 
     /// Scan a single `TxOut` for a matching script pubkey.
     ///
     /// If it matches the index will store and index it.
     pub fn scan_txout(&mut self, op: OutPoint, txout: &TxOut) {
-        self.inner.scan_txout(op, &txout);
+        self.inner.scan_txout(op, &txout)
     }
 
     pub fn inner(&self) -> &SpkTxOutIndex<(K, u32)> {
@@ -144,8 +145,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         self.inner
             .script_pubkeys()
             .range(&(keychain.clone(), u32::MIN)..=&(keychain.clone(), u32::MAX))
+            .next_back()
             .map(|((_, index), _)| *index)
-            .last()
     }
 
     /// Gets the current derivation index for each keychain in the index.
@@ -280,6 +281,41 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
                     .map(|index| (keychain.clone(), index))
             })
             .collect()
+    }
+
+    pub fn derive_until_unused_gap(&mut self, gap: u32) -> bool {
+        if gap == 0 {
+            return false;
+        }
+
+        let up_to_per_keychain = self
+            .keychains()
+            .iter()
+            .map(|(keychain, _)| {
+                let up_to = self
+                    .last_active_index(keychain)
+                    .unwrap_or(gap.saturating_sub(1));
+                (keychain.clone(), up_to)
+            })
+            .collect::<BTreeMap<_, _>>();
+
+        self.store_all_up_to(&up_to_per_keychain)
+    }
+
+    pub fn prune_unused(&mut self, mut keychain_bounds: BTreeMap<K, u32>) {
+        for (keychain, _) in &self.keychains {
+            keychain_bounds.entry(keychain.clone()).or_insert(0);
+        }
+        for (keychain, lower_bound) in keychain_bounds {
+            let indexes_to_prune = self
+                .inner
+                .unused((keychain.clone(), lower_bound)..=(keychain.clone(), u32::MAX))
+                .map(|((_, i), _)| *i)
+                .collect::<Vec<_>>();
+            for index in indexes_to_prune {
+                self.inner.remove_unused(&(keychain.clone(), index));
+            }
+        }
     }
 }
 
--- a/bdk_keychain/src/keychain_txout_index.rs
+++ b/bdk_keychain/src/keychain_txout_index.rs
@@ -166,7 +166,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     pub fn store_all_up_to(&mut self, keychains: &BTreeMap<K, u32>) -> bool {
         keychains
             .into_iter()
-            .any(|(keychain, index)| self.store_up_to(keychain, *index))
+            .map(|(keychain, index)| self.store_up_to(keychain, *index))
+            .fold(false, |acc, r| acc || r)
     }
 
     /// Derives script pubkeys from the descriptor **up to and including** `up_to` and stores them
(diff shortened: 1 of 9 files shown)


bdk_rpc_example/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+enum RpcCommands {
+    /// Scans blocks from bitcoin-core RPC and records relevant transactions
+    Scan {
+        /// Block height in which to start intial scan
+        #[clap(long, default_value = "0")]
+        initial_height: u64,
+        /// The unused-script gap will be kept at this value
+        #[clap(long, default_value = "5")]
+        stop_gap: usize,
+    },
```
LLFourn, line 53: What bout `Sync` which uses the last checkpoint as initial height and doesn't have a stop gap (just scans already derived scripts).

bdk_rpc_example/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+pub fn ensure_spk_cached<K>(txout_index: &mut KeychainTxOutIndex<K>, stop_gap: usize)
+where
+    K: Ord + Clone + core::fmt::Debug,
+{
+    txout_index.trailing_unused_counts().into_iter().for_each(
+        |(keychain, trailing_unused_count)| {
+            let missing_count = stop_gap.saturating_sub(trailing_unused_count);
+            txout_index.derive_new_count(&keychain, missing_count);
+        },
+    );
```
LLFourn, line 181: I think this should be simpler:

```rust
// for each keychain
let store_up_to = txout_index.last_active_index(&keychain).map(|lai| lai + stop_gap).unwrap_or(stop_gap);
txout_index.store_up_to(derive_up_to);
```

It would be fine to add a helper method to `KeychainTxOutIndex` to make sure all it's keychains have stored up to `last_active_index + stop_gap`.

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -257,6 +266,28 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
         }
     }
+
+    pub fn trailing_unused_count(&self, keychain: &K) -> usize {
+        let last_index = self
+            .inner
+            .range_unused((keychain.clone(), u32::MIN)..=(keychain.clone(), u32::MAX))
+            .last()
+            .map(|(&(_, index), _)| index)
+            .unwrap_or(0);
```
LLFourn, line 276: Related to below. I think it's more useful and intuitive to have provide `last_active_index(&self, keychain: &K) -> Option<u32>`. Trailing unused is `derivation_index` - `last_active_index`.

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -233,6 +233,15 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
     }
 
+    pub fn derive_new_count(&mut self, keychain: &K, count: usize) -> BTreeMap<u32, Script> {
+        let mut out = BTreeMap::new();
+        for _ in 0..count {
+            let (index, script) = self.derive_new(keychain);
+            out.insert(index, script.clone());
+        }
+        out
+    }
```
LLFourn, line 243: what's wrong with just using `store_up_to` where you pass in the index you want to get to. You can translate any call to `derive_new_count` to `store_up_to` can't you?

Not sure what's going to be done with return value here.

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -78,15 +78,15 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// See [`ForEachTxout`] for the types that support this.
     ///
     /// [`ForEachTxout`]: bdk_core::ForEachTxout
-    pub fn scan(&mut self, txouts: &impl ForEachTxout) {
-        self.inner.scan(txouts);
+    pub fn scan(&mut self, txouts: &impl ForEachTxout) -> BTreeSet<(K, u32)> {
+        self.inner.scan(txouts)
```
LLFourn, line 82: why does this return value need to exist?

bdk_rpc_example/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+
+                    for (keychain, this_ind) in keychain_tracker.txout_index.scan(&tx) {
+                        last_active_indexes
+                            .entry(keychain)
+                            .and_modify(|ind| {
+                                if this_ind > *ind {
+                                    *ind = this_ind;
+                                }
+                            })
+                            .or_insert(this_ind);
```
LLFourn, line 149: Like I've been saying I think you should just have a method on `txout_index` to get this value. Very useful method in general that just happens to be missing.

LLFourn (changes requested):
Thanks @evanlinjin. Quick review of how bdk core APIs are used. I think it can be simplified quite a bit by changing the API to return `last_active_index`.

I would love to see the sync/scan command to have a `--live` option where after doing the sync it prints out some basic stuff like balance and utxos maybe and then sits there polling for new blocks and prints out details about new transactions that come in that spend/create new wallet utxos.

Since the codebase is getting more mature now it'd be great if you could add docs and even example usage for new APIs (or even old ones). I find writing an explanation of the API helps me figure out early if the API actually makes sense.

bdk_rpc_example/src/main.rs
evanlinjin: I currently don't think RPC requires a difference between sync/scan. Either way we still need to look through every block, starting from the "point of agreement". Stop gap can be set to 0 if we are good to work with current cached scripts.

Let me know if I am missing something.

LLFourn: Ahh right. Then we shouldn't call this "stop_gap" since we don't stop if we find a gap right? "address_look_ahead" or something.

bdk_rpc_example/src/main.rs
```diff
@@ -0,0 +1,240 @@
...
+                };
+
+                for (height, tx) in txs {
+                    keychain_tracker
+                        .txout_index
+                        .derive_until_unused_gap(stop_gap);
+                    if keychain_tracker.txout_index.is_relevant(&tx) {
+                        update.insert_tx(tx.clone(), Some(height))?;
+                    }
+                    keychain_tracker.txout_index.scan(&tx);
```
LLFourn, line 168: I think this is unsound -- you have to scan the txs first before deciding if it's relevant. scan the block when it comes in and each tx individually when the mempool comes in. From the docs of `is_relevant`:

```
    /// Whether any of the inputs of this transaction spend a txout tracked or whether any output
    /// matches one of our script pubkeys.
    ///
    /// It is easily possible to misuse this method and get false negatives by calling it before you
    /// have scanned the `TxOut`s the transaction is spending. For example if you want to filter out
    /// all the transactions in a block that are irrelevant you **must first scan all the
    /// transactions in the block** and only then use this method.
```

evanlinjin: But we are scanning transactions in order so the previous outputs will be scanned before the current transaction right?

evanlinjin:
@LLFourn I'm not a fan of having changesets for last_used_indicies

bdk_rpc_example/src/main.rs
```diff
@@ -0,0 +1,182 @@
...
+                }
+
+                let wallet_scan = KeychainScan {
+                    update,
+                    last_active_indexes,
+                };
+
+                let changeset = keychain_tracker.determine_changeset(&wallet_scan)?;
+                db.append_changeset(&changeset)?;
+                keychain_tracker.apply_changeset(changeset);
```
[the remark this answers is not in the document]
evanlinjin: > I see you've avoided the need to "trim" the spks derived by only storing the last active indexes in the database. Can we add that anyway? i.e. imagine that you need to keep using the keychain_tracker after this i.e. get new address. We want it to return the one right after the last active index.

This is now done here:

https://github.com/LLFourn/bdk_core_staging/blob/80e116a27afc1653c5a920778754a98d2bbff866/bdk_keychain/src/keychain_txout_index.rs#L305-L320

> A good motivating feature would be to create a REPL mode for the cli where it does a scan and then puts you into a REPL where you can call all the cli stuff but it will also poll for new blocks/mempool txs in the background.

This will be great. Let's do that in a separate PR? #93

bdk_rpc_example/src/main.rs
LLFourn: Oh that's a good point but that only applies to the block txs. In the mempool are you guaranteed some kind of logical traversal order when you get the response. Maybe? But we shouldn't depend on it.

evanlinjin:
Replaced by #157

All of @LLFourn 's pointers here should be addressed.

PR #92, opened 2022-12-19
Make chain_graph::apply_changeset safe

LLFourn:
- Make ChainGraph::apply_changeset safe (doesn't add txid to chain without adding full tx to graph).
- Add inflate_changeset to safely turn a sparse_chain changeset into a chain_graph changeset.

This is to make #77 a bit cleaner. See PR to #77 here: https://github.com/rajarshimaitra/bdk_core_staging/pull/1 which does the best job we can do without these changes.

### DIFF
--- a/bdk_core/src/chain_graph.rs
+++ b/bdk_core/src/chain_graph.rs
@@ -1,4 +1,5 @@
 use crate::{
+    collections::HashSet,
     sparse_chain::{self, ChainIndex, SparseChain},
     tx_graph::{self, TxGraph},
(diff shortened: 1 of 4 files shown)


evanlinjin:
Will review tomorrow morning

LLFourn:
@evanlinjin good point. I made an issue for this: https://github.com/LLFourn/bdk_core_staging/issues/97 I think we should merge this now and do #97 soon. We are already not consistent in applying those rules e.g. `insert_tx` doesn't check.

evanlinjin (approved):

PR #95, opened 2022-12-20
Llfourn rpc example

LLFourn:
This is a PR just to show something I tried out to @evanlinjin. Feel free to just steal any good ideas from it and drop it.


evanlinjin:
@LLFourn Thanks for this PR! I've looked through, and will incorporate some of these simplifications into #157 

Closing for now.

PR #100, opened 2022-12-21
Add ChainGraph::get_tx_in_chain

LLFourn:
A useful utility for getting a full transaction in the chain.

Also deletes bincode feature flag not used that I just noticed.


evanlinjin (approved):
ACK 4fd78d47bd0c348834772ec577371bcba3f8159c

PR #103, opened 2022-12-21
Allow adding custom cli args

evanlinjin:
Also added `Args::cp_limit`.

### DIFF
--- a/bdk_cli_lib/src/lib.rs
+++ b/bdk_cli_lib/src/lib.rs
@@ -24,7 +24,7 @@ use std::{cmp::Reverse, collections::HashMap, fmt::Debug, path::PathBuf, time::D
 #[derive(Parser)]
 #[clap(author, version, about, long_about = None)]
 #[clap(propagate_version = true)]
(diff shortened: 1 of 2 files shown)


rajarshimaitra (approved):
ACK 8e9395902d21576a531516a0ef28b61a7d7470ff

LLFourn:
ACK though in practice I think we should not put rpc args at the top level. The rpc args should be put on the actual command that uses them. Otherwise, It strongly implies to someone that every command needs these arguments.

evanlinjin:
> ACK though in practice I think we should not put rpc args at the top level. The rpc args should be put on the actual command that uses them. Otherwise, It strongly implies to someone that every command needs these arguments.

What do you mean by RPC Args?

LLFourn:
@evanlinjin I mean that it should be an argument to the subcommand only.

evanlinjin:
As discussed with @LLFourn, I am convinced that `chain_args` is not needed. However, `cp_limit` will still be helpful.

PR #104, opened 2022-12-21
Add methods to prune `KeychainTxOutIndex`

evanlinjin:
Also:

* Fix `KeychainTxOutIndex::store_all_up_to`.
* Make `KeychainTxOutIndex::derivation_index` more efficient.

### DIFF
--- a/bdk_core/src/spk_txout_index.rs
+++ b/bdk_core/src/spk_txout_index.rs
@@ -191,6 +191,16 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
         self.unused.get(index).is_none()
     }
 
+    /// Removes an unused index.
+    pub fn remove_unused(&mut self, index: &I) -> bool {
+        let is_removed = self.unused.remove(index);
+        if is_removed {
+            let spk = self.script_pubkeys.remove(index).expect("should exist");
+            self.spk_indexes.remove(&spk);
+        }
+        is_removed
+    }
+
     /// Returns the index associated with the script pubkey.
     pub fn index_of_spk(&self, script: &Script) -> Option<I> {
         self.spk_indexes.get(script).cloned()
--- a/bdk_keychain/src/keychain_txout_index.rs
+++ b/bdk_keychain/src/keychain_txout_index.rs
@@ -1,3 +1,4 @@
+use alloc::vec::Vec;
 use bdk_core::{
     bitcoin::{secp256k1::Secp256k1, OutPoint, Script, TxOut},
     collections::*,
@@ -246,6 +247,61 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         } else {
             self.keychain_unused(keychain).next().unwrap()
         }
+    }
+
+    /// Convenience method to call [`store_until_unused_count`] for all keychains.
+    ///
+    /// [`store_until_unused_count`]: Self::store_until_unused_count
+    pub fn store_all_until_unused_count(&mut self, unused_count: u32) -> bool {
+        let keychains = self.keychains.keys().cloned().collect::<Vec<_>>();
+        keychains
+            .into_iter()
+            .map(|keychain| self.store_until_unused_count(&keychain, unused_count))
+            .fold(false, |acc, v| acc || v)
+    }
+
+    /// Derives script pubkeys from the descriptor until the unused script count reaches
+    /// `unused_count`.
+    ///
+    /// Returns whether any new script pubkeys were derived.
+    pub fn store_until_unused_count(&mut self, keychain: &K, unused_count: u32) -> bool {
+        if unused_count < 1 {
+            return false;
+        }
+        let up_to = self
+            .last_active_index(keychain)
+            .map(|i| i.saturating_add(unused_count))
+            .unwrap_or(unused_count.saturating_sub(1));
+        self.store_up_to(keychain, up_to)
+    }
+
+    /// Convenience method to call [`prune_unused`] for all keychains.
+    ///
+    /// `prune_from` contains the lowest allowed index to prune from per keychain (default is 0).
+    ///
+    /// [`prune_unused`]: Self::prune_unused
+    pub fn prune_all_unused(&mut self, prune_from: BTreeMap<K, u32>) -> bool {
+        let keychains = self.keychains.keys().cloned().collect::<Vec<_>>();
+        keychains
+            .into_iter()
+            .map(|keychain| self.prune_unused(&keychain, *prune_from.get(&keychain).unwrap_or(&0)))
+            .fold(false, |acc, v| acc || v)
+    }
+
+    /// Prunes unused script pubkeys of a given keychain, starting from index `prune_from`.
+    ///
+    /// Returns whether anything was removed.
+    pub fn prune_unused(&mut self, keychain: &K, prune_from: u32) -> bool {
+        let to_prune = self
+            .inner
+            .unused((keychain.clone(), prune_from)..=(keychain.clone(), u32::MAX))
+            .map(|(index, _)| index.clone())
+            .collect::<Vec<_>>();
+
+        to_prune
+            .into_iter()
+            .map(|index| self.inner.remove_unused(&index))
+            .fold(false, |acc, v| acc || v)
     }
 
     /// Iterates over all unused script pubkeys for a `keychain` that have been stored in the index.
--- a/bdk_keychain/src/keychain_txout_index.rs
+++ b/bdk_keychain/src/keychain_txout_index.rs
@@ -145,8 +145,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         self.inner
             .script_pubkeys()
             .range(&(keychain.clone(), u32::MIN)..=&(keychain.clone(), u32::MAX))
+            .next_back()
             .map(|((_, index), _)| *index)
-            .last()
     }
 
     /// Gets the current derivation index for each keychain in the index.
@@ -166,7 +166,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     pub fn store_all_up_to(&mut self, keychains: &BTreeMap<K, u32>) -> bool {
         keychains
             .into_iter()
-            .any(|(keychain, index)| self.store_up_to(keychain, *index))
+            .map(|(keychain, index)| self.store_up_to(keychain, *index))
+            .fold(false, |acc, v| acc || v)
     }
 
     /// Derives script pubkeys from the descriptor **up to and including** `up_to` and stores them


bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -248,6 +250,61 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+    ///
+    /// Returns whether any new script pubkeys were derived.
+    pub fn store_until_unused_count(&mut self, keychain: &K, unused_count: u32) -> bool {
+        if unused_count < 1 {
+            return false;
+        }
+        let up_to = self
+            .last_active_index(keychain)
+            .map(|i| i.saturating_add(unused_count))
+            .unwrap_or(unused_count.saturating_sub(1));
```
rajarshimaitra, line 275: This did not made sense to me.. Shouldn't the default in this case be `u32::MAX`?

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -248,6 +250,61 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+    /// Convenience method to call [`prune_unused`] for all keychains.
+    ///
+    /// `prune_from` contains the lowest allowed index to prune from per keychain (default is 0).
+    ///
+    /// [`prune_unused`]: Self::prune_unused
+    pub fn prune_all_unused(&mut self, prune_from: BTreeMap<K, u32>) -> bool {
+        let keychains = self.keychains.keys().cloned().collect::<Vec<_>>();
+        keychains
+            .into_iter()
+            .map(|keychain| self.prune_unused(&keychain, *prune_from.get(&keychain).unwrap_or(&0)))
```
rajarshimaitra, line 288: Umm.. It seems this will prune everything if the `keychain` isn't present in the map and `get` returns `None`??

rajarshimaitra (commented):
I did an initial review and have few questions below. Also when is removing unused scripts useful?

bdk_keychain/src/keychain_txout_index.rs
rajarshimaitra, line 288: This looks like it will remove everything from a keychain if that's not include in the map and `get` returns `None`??

Instead of `self.keychains` I think this should iterate over the `prune_from` map only..

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -248,6 +250,61 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+            .fold(false, |acc, v| acc || v)
+    }
+
+    /// Prunes unused script pubkeys of a given keychain, starting from index `prune_from`.
+    ///
+    /// Returns whether anything was removed.
+    pub fn prune_unused(&mut self, keychain: &K, prune_from: u32) -> bool {
```
LLFourn, line 295: Let's call this `trim_unused` since trim is already used like this in `std`. I think we don't want the `prune_from` parameter. It should just prune to the last active index. The reason you want this is you are trying to restore derivation indices to what they were before right? But we should not be doing this restoration. There are three cases:

1. We are doing an initial scan after restoring from seed words. In this case we don't care about what the derivation indices were before we started because they are all 0.
2. We are updating our wallet with a new block. In this case we don't need to trim since we don't need to do pad with unused since we are the authoritative source on our derivation indices.
3. We are a watch only wallet. In this case we will pad with unused but we never need to trim.

In none of the cases do we need this parameter. This is why I think you need two commands in your rpc example: `Scan` and `Sync` which are scenarios (1) and (2) respectively.

bdk_core/src/spk_txout_index.rs
```diff
@@ -191,6 +191,16 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
         self.unused.get(index).is_none()
     }
 
+    /// Removes an unused index.
```
LLFourn, line 194: document return value.

bdk_core/src/spk_txout_index.rs
```diff
@@ -191,6 +191,16 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
...
+    pub fn remove_unused(&mut self, index: &I) -> bool {
+        let is_removed = self.unused.remove(index);
+        if is_removed {
+            let spk = self.script_pubkeys.remove(index).expect("should exist");
```
LLFourn, line 198:
```suggestion
            let spk = self.script_pubkeys.remove(index).expect("must exist");
```

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -165,7 +166,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     pub fn store_all_up_to(&mut self, keychains: &BTreeMap<K, u32>) -> bool {
         keychains
             .into_iter()
-            .any(|(keychain, index)| self.store_up_to(keychain, *index))
+            .map(|(keychain, index)| self.store_up_to(keychain, *index))
+            .fold(false, |acc, v| acc || v)
```
LLFourn, line 170: Thanks for finding this. I think it would be better style to do this is with a for loop and a bool variable.

LLFourn (changes requested):
Hey @evanlinjin. Thanks for this. It was worth pulling these out into their own PR. I made some potentially contentious suggestions. If you agree with them, please go ahead and add some tests for the new methods. It'd be good to have some to cover the bug you found and the bug (I think) I found in your new function.

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -144,8 +145,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         self.inner
             .script_pubkeys()
             .range(&(keychain.clone(), u32::MIN)..=&(keychain.clone(), u32::MAX))
+            .next_back()
```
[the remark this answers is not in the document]
evanlinjin: @LLFourn Thanks I did not know that, will change it back!

bdk_keychain/src/keychain_txout_index.rs
```diff
@@ -248,6 +250,61 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+    /// Derives script pubkeys from the descriptor until the unused script count reaches
+    /// `unused_count`.
+    ///
+    /// Returns whether any new script pubkeys were derived.
+    pub fn store_until_unused_count(&mut self, keychain: &K, unused_count: u32) -> bool {
+        if unused_count < 1 {
+            return false;
+        }
+        let up_to = self
+            .last_active_index(keychain)
```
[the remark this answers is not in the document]
evanlinjin: I like the term `pad` for the naming.

bdk_keychain/src/keychain_txout_index.rs
evanlinjin: Not "everything", it will remove unused. `prune_from` sets where specifically to prune from, if it is not set for a keychain, it will prune all unused (all in our new terminology, prune unused "padding").

If the user would like to prune for specific keychains, they can just use `prune_unused`.

rajarshimaitra: Okay, so the intention here is to prune all unused for keychains if their `prune_from` isn't set.. Ya that makes sense.. Maybe it can be included in the doc to make this intention a bitcoin clear.. I was thinking more like if there are 5 keychains, and I only wanna prune 3 of them, then I would use this.. But that's definitely not the case here..

bdk_keychain/src/keychain_txout_index.rs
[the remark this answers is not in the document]
rajarshimaitra: @LLFourn Thanks ya that makes sense, I missed that point..

bdk_keychain/src/keychain_txout_index.rs
rajarshimaitra: I like this style tbh.. 😄 Quite cool and compact..

bdk_keychain/src/keychain_txout_index.rs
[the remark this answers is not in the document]
rajarshimaitra: > We could add a way of enforcing this which could be handy i.e. derive_new returns an Option<(u32, &Script)> that is None when there are already too many unused ones. Out of scope for this PR though.

LMK if I am missing something, but it seems the `derive_next_unused` already maintaining this? Maybe we should call it `get_next_unused` because it only derives when there are no unused left. I think that solves this purpose, and we should not get a new unused when there are already unused left. And `derive_new` can stay as it is, as that's specifically asking to derive new keys.

bdk_keychain/src/keychain_txout_index.rs
rajarshimaitra: Ah okay.. I was confusing this `None` with `saturating_add` return, I don't know why.. Ya now it makes sense..

bdk_keychain/src/keychain_txout_index.rs
LLFourn: It was even more cool and compact before and it was a bug.

bdk_keychain/src/keychain_txout_index.rs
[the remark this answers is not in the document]
LLFourn: Added issue: https://github.com/LLFourn/bdk_core_staging/issues/110

bdk_keychain/src/keychain_txout_index.rs
evanlinjin: Personally I don't think `Sync` is ever a good idea with our model. What if the user has two instances of the same descriptor (`WalletA` and `WalletB`) and the following scenario plays out:
* `WalletA` is currently synced up to height 100 and last_derivation_index is 10.
* `WalletB` receives $$$ via txs that are confirmed in block 101 and last_used_index is 12.
* After some time, `WalletA` initiates a `Sync` up to height 102. Our model will fail to find incoming txs sending to indexes 11 and 12.

Hence, I think it will be safer to always enforce a minimum lookahead count (which is scan).

LLFourn: > What if the user has two instances of the same descriptor (`WalletA` and `WalletB`) and the following scenario plays out:

You should not design a system like this because it results in address re-use. I'd say it's a feature if our API makes it really awkward to implement this since it will stop and make you think about other designs. Each system that gives out addresses should be the only system giving out addresses on that keychain. The rule is simple, if you are a system that gives out the addresses do a sync and make sure to manage derivation indices correctly, otherwise do a scan and just pad your keychain with a bunch of pre-derived addresses.

evanlinjin:
The only things that are left to do here are the "trimming" methods (everything else have been done in some way or another).

However, maybe we should have further discussion on whether these methods are actually useful.

evanlinjin:
Not needed anymore.

PR #106, opened 2022-12-22
Fix KeychainTracker::insert_tx

LLFourn:
I forgot to scan the output after inserting it.
I also added some natural `add_keychain` and `keychains` methods on it.


evanlinjin (approved):
ACK 1771992cf31a0f8c1ac705d770e285afc4fc191b

PR #108, opened 2022-12-23
Add SpkTxOutIndex::sent_and_received

LLFourn:
and add net_value while I'm at it.


evanlinjin (approved):
ACK a2b29a953fe48482074ffdc388c364df282b1aa0

Issue #97, opened 2022-12-20
Enforce tx consistency in ChainGraph


LLFourn:
Now that we have done #80 we always have a full tx whenever we have a txid in the `SparseChain`. In #74  we started enforcing that txs in `ChainGraph` don't double spend each other but only in `determine_changeset`. We are inconsistent about this though. We don't do it in `insert_tx` for example. This is an issue to discuss whether we want to do this across the whole structure. If we are not going to always enforcing it then we shouldn't just do it in `determine_changeset`. We should have a separate method like `determine_changeset_evict_conflicts` that does `determine_changeset`.

My preference is to keep it consistent across the whole structure.

rajarshimaitra:
> My preference is to keep it consistent across the whole structure.

+1 on this.. This makes reasoning over the structure easier and less chance to screw up by the API.. IMO we should maintain all the invariant consistently across the API..

evanlinjin:
> We don't do it in `insert_tx` for example.

However we also don't do it for `apply_update` as well.

How should we approach this? For `insert_tx`, should we split it into `determine_insert_tx_changeset` and `apply_insert_tx`?

PR #109, opened 2022-12-23
Fix conflicts in inflate_changeset

LLFourn:


LLFourn:
I think this should be carried through all the way to fix #97

evanlinjin:
> I think this should be carried through all the way to fix #97

As discussed, we should defer #97

evanlinjin (approved):

PR #116, opened 2022-12-24
`insert..` methods ensure chain consistency + renaming stuff

### LINKED ISSUE #90: Remove "iter_" prefix for things
In `std` many things return an iterator without putting `iter_` in front of it. Let's only put `iter` when we are creating some kind of custom iterator (e.g. generating a iterator for all spks of a descriptor) and want to bring attention to that.

### LINKED ISSUE #97: Enforce tx consistency in ChainGraph

### LINKED ISSUE #110: Rename derive_next_unused next_unused
Since it doesn't always derive. 

see https://github.com/LLFourn/bdk_core_staging/pull/104#discussion_r1055125722

evanlinjin:
Closes #97 
Closes #90 
Closes #110 

**Chain consistency stuff**

Methods such as `insert_{}` are split between `insert_{}_preview` and `insert_{}`. One only previews the changes, the latter also applies. Both these methods returns a changeset.

The internal logic of `insert_{}_preview` now uses as `determine_changeset` to make behavior more reliable and reduce duplicated code. This means that any changes are consistent with the logic of `determine_changeset/additions`.

**Renaming stuff**

* Renamed `KeychainTxOutIndex::derive_next_unused` to `next_unused`.
* Remove `iter_` prefixes from methods.

### DIFF
--- a/bdk_chain/src/chain_graph.rs
+++ b/bdk_chain/src/chain_graph.rs
@@ -287,7 +287,7 @@ impl<P: ChainPosition> ChainGraph<P> {
     /// in ascending order.
     pub fn transactions_in_chain(&self) -> impl DoubleEndedIterator<Item = (&P, &Transaction)> {
         self.chain
-            .iter_txids()
+            .txids()
             .map(|(pos, txid)| (pos, self.graph.tx(*txid).expect("must exist")))
     }
 
--- a/bdk_chain/src/chain_graph.rs
+++ b/bdk_chain/src/chain_graph.rs
@@ -4,6 +4,7 @@ use crate::{
     tx_graph::{self, TxGraph},
     BlockId, ForEachTxout, FullTxOut, TxHeight,
 };
+use alloc::collections::BTreeMap;
 use bitcoin::{OutPoint, Transaction, TxOut, Txid};
 use core::fmt::Debug;
 
@@ -55,13 +56,27 @@ impl<P: ChainPosition> ChainGraph<P> {
     /// Inserts a transaction into the inner [`ChainGraph`] and optionally into the chain at
     /// `position`.
     ///
+    /// If inserting it into a chain `position` will result in a conflict, an error will be returned
+    /// instead.
+    ///
     /// **Warning**: This function modifies the internal state of the chain graph. You are
     /// responsible for persisting these changes to disk if you need to restore them.
     pub fn insert_tx(
         &mut self,
         tx: Transaction,
         position: Option<P>,
-    ) -> Result<bool, sparse_chain::InsertTxErr> {
+    ) -> Result<bool, InsertTxErr<P>> {
+        // only care about conflicts if chain position is provided
+        if position.is_some() {
+            let conflicts = self
+                .conflicting_txids(&tx)
+                .map(|(k, v)| (k.clone(), v))
+                .collect::<BTreeMap<_, _>>();
+            if !conflicts.is_empty() {
+                return Err(InsertTxErr::Conflicts(conflicts));
+            }
+        }
+
         let chain_changed = match position {
             Some(pos) => self.chain.insert_tx(tx.txid(), pos)?,
             None => false,
@@ -116,6 +131,21 @@ impl<P: ChainPosition> ChainGraph<P> {
         Ok(changeset)
     }
 
+    /// Given a transaction, return an iterator of conflicting [`Txid`]s that are marked within the
+    /// internal chain.
+    pub fn conflicting_txids<'a>(
+        &'a self,
+        tx: &'a Transaction,
+    ) -> impl Iterator<Item = (&'a P, Txid)> + 'a {
+        self.graph
+            .conflicting_txids(tx)
+            .filter_map(|(_, conflicting_txid)| {
+                self.chain
+                    .tx_position(conflicting_txid)
+                    .map(|conflicting_pos| (conflicting_pos, conflicting_txid))
+            })
+    }
+
     fn fix_conflicts(&self, changeset: &mut ChangeSet<P>) -> Result<(), UnresolvableConflict<P>> {
         let chain_conflicts = changeset
             .graph
@@ -125,21 +155,16 @@ impl<P: ChainPosition> ChainGraph<P> {
             // we care about transactions that are not already in the chain
             .filter(|(_, txid)| self.chain.tx_position(*txid).is_none())
             // and are going to be added to the chain
-            .filter_map(|(tx, txid)| Some((changeset.chain.txids.get(&txid)?.clone()?, txid, tx)))
+            .filter_map(|(tx, txid)| Some((changeset.chain.txids.get(&txid)?.as_ref()?, txid, tx)))
             .flat_map(|(update_pos, update_txid, update_tx)| {
-                self.graph
-                    .conflicting_txids(update_tx)
-                    .filter_map(move |(_, conflicting_txid)| {
-                        self.chain
-                            .tx_position(conflicting_txid)
-                            .map(|conflicting_pos| {
-                                (
-                                    update_pos.clone(),
-                                    update_txid,
-                                    conflicting_pos,
-                                    conflicting_txid,
-                                )
-                            })
+                self.conflicting_txids(update_tx)
+                    .map(move |(conflicting_pos, conflicting_txid)| {
+                        (
+                            update_pos.clone(),
+                            update_txid,
+                            conflicting_pos,
+                            conflicting_txid,
+                        )
                     })
             })
             .collect::<alloc::vec::Vec<_>>();
@@ -150,14 +175,14 @@ impl<P: ChainPosition> ChainGraph<P> {
             // If so, we will modify the changeset to evict the conflicting txid.
 
             // determine the position of the conflicting txid after current changeset is applied
-            let conflicting_new_ind = changeset
+            let conflicting_new_pos = changeset
                 .chain
                 .txids
                 .get(&conflicting_txid)
                 .map(Option::as_ref)
                 .unwrap_or(Some(conflicting_pos));
 
-            match conflicting_new_ind {
+            match conflicting_new_pos {
                 None => {
                     // conflicting txid will be deleted, can ignore
                 }
@@ -167,7 +192,7 @@ impl<P: ChainPosition> ChainGraph<P> {
                         // evicted, return error
                         return Err(UnresolvableConflict {
                             already_confirmed_tx: (conflicting_pos.clone(), conflicting_txid),
-                            update_tx: (update_pos, update_txid),
+                            update_tx: (update_pos.clone(), update_txid),
                         });
                     }
                     TxHeight::Unconfirmed => {
@@ -311,6 +336,33 @@ impl<P> ForEachTxout for ChangeSet<P> {
         self.graph.for_each_txout(f)
     }
 }
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum InsertTxErr<P> {
+    Chain(sparse_chain::InsertTxErr),
+    Conflicts(BTreeMap<P, Txid>),
+}
+
+impl<P: core::fmt::Debug> core::fmt::Display for InsertTxErr<P> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            InsertTxErr::Chain(inner) => core::fmt::Display::fmt(inner, f),
+            InsertTxErr::Conflicts(_) => write!(
+                f,
+                "cannot directly insert conflicting transaction into chaingraph"
+            ),
+        }
+    }
+}
+
+impl<P> From<sparse_chain::InsertTxErr> for InsertTxErr<P> {
+    fn from(inner: sparse_chain::InsertTxErr) -> Self {
+        Self::Chain(inner)
+    }
+}
+
+#[cfg(feature = "std")]
+impl<P: core::fmt::Debug> std::error::Error for InsertTxErr<P> {}
 
 /// Represents an update failure.
 #[derive(Clone, Debug, PartialEq)]
--- a/bdk_chain/src/chain_graph.rs
+++ b/bdk_chain/src/chain_graph.rs
@@ -4,7 +4,6 @@ use crate::{
     tx_graph::{self, TxGraph},
     BlockId, ForEachTxout, FullTxOut, TxHeight,
 };
-use alloc::collections::BTreeMap;
 use bitcoin::{OutPoint, Transaction, TxOut, Txid};
 use core::fmt::Debug;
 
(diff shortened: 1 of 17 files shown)


LLFourn (changes requested):
LGTM! Tests please. I have a noted a few issues that have to be opened up. I'll do that when I get a moment.

bdk_chain/src/chain_graph.rs
```diff
@@ -55,13 +56,27 @@ impl<P: ChainPosition> ChainGraph<P> {
...
     ///
+    /// If inserting it into a chain `position` will result in a conflict, an error will be returned
+    /// instead.
+    ///
     /// **Warning**: This function modifies the internal state of the chain graph. You are
     /// responsible for persisting these changes to disk if you need to restore them.
     pub fn insert_tx(
         &mut self,
         tx: Transaction,
         position: Option<P>,
```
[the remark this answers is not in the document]
evanlinjin: > For now I'd say just remove the ability to insert things into the underlying graph without a position (at the point of insertion).

I feel like the ability to insert arbitrary things into the underlying graph is a good feature to have. For example, a wallet may want to track all "replaced" transactions. Returning a mutable reference to the `TxGraph` makes sense to me, is it alright if I include this in the same PR here?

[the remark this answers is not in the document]
LLFourn: > For example, a wallet may want to track all "replaced" transactions

For this use case you can insert them as unconfirmed before they are replaced i.e. insert them in order. They will still be in the graph even if they are evicted from the chain. For now do not expose the &mut ChainGraph I will explain in an issue shortly.

bdk_chain/src/chain_graph.rs
```diff
@@ -55,13 +56,27 @@ impl<P: ChainPosition> ChainGraph<P> {
...
-    ) -> Result<bool, sparse_chain::InsertTxErr> {
+    ) -> Result<bool, InsertTxErr<P>> {
+        // only care about conflicts if chain position is provided
+        if position.is_some() {
```
rajarshimaitra, line 70: Why do we enforce this if position of the new tx isn't required to find conflicts?

bdk_chain/src/chain_graph.rs
```diff
@@ -55,13 +56,27 @@ impl<P: ChainPosition> ChainGraph<P> {
...
+                .map(|(k, v)| (k.clone(), v))
+                .collect::<BTreeMap<_, _>>();
+            if !conflicts.is_empty() {
+                return Err(InsertTxErr::Conflicts(conflicts));
+            }
+        }
+
         let chain_changed = match position {
             Some(pos) => self.chain.insert_tx(tx.txid(), pos)?,
             None => false,
```
rajarshimaitra, line 82: To me it seems if `position` is `None`, we will add it only to graph and not in chain. Is that intended? And then we can add conflicting transactions too in the graph, by bypassing the filter above?

bdk_chain/src/chain_graph.rs
```diff
@@ -116,6 +131,21 @@ impl<P: ChainPosition> ChainGraph<P> {
...
+        tx: &'a Transaction,
+    ) -> impl Iterator<Item = (&'a P, Txid)> + 'a {
+        self.graph
+            .conflicting_txids(tx)
+            .filter_map(|(_, conflicting_txid)| {
+                self.chain
+                    .tx_position(conflicting_txid)
+                    .map(|conflicting_pos| (conflicting_pos, conflicting_txid))
+            })
+    }
```
rajarshimaitra, line 147: I might be confused, but are we saying, we will only find conflicts if they are both in graph and chain? What happens for conflicts which are **only in graph** and **not in chain**? if we don't detect them, should we clarify that in the comment?

rajarshimaitra (commented):
Made a basic review of caabad49fdfbd2f85d082ec2df2c120a4c6eb5de

I understand this is currently being worked on and would do another pass after next changes. Mostly I just have few questions for now.

bdk_chain/src/chain_graph.rs
LLFourn: We don't care about conflicts that are in the graph but not the chain. The chain is the thing that must be consistent. This is a crucial concept to understand when using the APIs and it should be very well documented but we don't need comments all over the place every time we employ this fact.

bdk_chain/src/chain_graph.rs
evanlinjin: Sorry, I'm not sure I know exactly what you mean. Do you mind expanding?

bdk_chain/src/chain_graph.rs
```diff
@@ -55,13 +56,27 @@ impl<P: ChainPosition> ChainGraph<P> {
...
+        if position.is_some() {
+            let conflicts = self
+                .conflicting_txids(&tx)
+                .map(|(k, v)| (k.clone(), v))
+                .collect::<BTreeMap<_, _>>();
+            if !conflicts.is_empty() {
+                return Err(InsertTxErr::Conflicts(conflicts));
+            }
+        }
+
```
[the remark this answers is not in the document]
evanlinjin: I've decided to have `insert_tx_changeset()` methods for `SparseChain`, `ChainGraph` etc. The returned changeset displays the required changes in order to insert the given tx (including evictions). Given the changeset, the user can decide whether or not to apply it.

bdk_chain/tests/test_tx_graph.rs
```diff
@@ -1,70 +1,67 @@
...
+//             tx: [].into(),
+//             txout: update_ops.into(),
+//         }
+//     );
+
+//     graph.apply_additions(additions);
+//     assert_eq!(graph.all_txouts().count(), 3);
+//     assert_eq!(graph.full_transactions().count(), 0);
+//     assert_eq!(graph.partial_transactions().count(), 2);
+// }
```
LLFourn, line 64: why is this test deleted?

bdk_chain/src/tx_graph.rs
```diff
@@ -104,15 +132,15 @@ impl TxGraph {
             .map(|txin| txin.previous_output)
             // coinbase spends are not to be counted
             .filter(|outpoint| !outpoint.is_null())
+            // record spends
```
LLFourn, line 135: :x: comment explains what not why.

bdk_chain/src/chain_graph.rs
```diff
@@ -52,25 +52,6 @@ impl<P: ChainPosition> ChainGraph<P> {
...
-    ) -> Result<bool, sparse_chain::InsertTxErr> {
-        let chain_changed = match position {
-            Some(pos) => self.chain.insert_tx(tx.txid(), pos)?,
-            None => false,
-        };
-        let graph_changed = self.graph.insert_tx(tx);
-
-        Ok(graph_changed || chain_changed)
-    }
-
```
LLFourn, line 73: I think we should not delete this API. Only giving `insert_tx_changeset` means the user has to deal with an error case when generating it AND applying it. I think this policy is better:

1. Have a function which returns the changeset with non mut `self`.
2. Have a utility function that genrates **and** applies the changeset and returns the changeset (by cloneing). This one would be called `insert_tx` for example.
3. Have a function that applies the changeset.
4. Make all changeset types `[must_use]`.

Let's apply this policy to all methods on all changeset types.

bdk_chain/tests/test_keychain_txout_index.rs
```diff
@@ -24,6 +24,7 @@ fn init_txout_index() -> KeychainTxOutIndex<TestKeychain> {
 }
 
 #[test]
+#[cfg(feature = "miniscript")]
```
LLFourn, line 27: this whole file should have `#![cfg(feature = "miniscript")]` I think I also had to do that in one of my other PRs.

bdk_chain/tests/common/mod.rs
```diff
@@ -48,3 +50,54 @@ macro_rules! changeset {
...
+#[allow(unused)]
+pub fn chain_insert_checkpoint<P: bdk_chain::sparse_chain::ChainPosition>(
+    chain: &mut bdk_chain::sparse_chain::SparseChain<P>,
+    block_id: bdk_chain::BlockId,
+) {
+    let changeset = chain
+        .insert_checkpoint_changeset(block_id)
+        .expect("must insert tx");
+    chain.apply_changeset(changeset);
+}
```
LLFourn, line 103: These functions seem to exist because the API currently has a lot of friction because of the two-step changeset. If it's worth getting rid of this friction while testing it's probably worth getting rid of it for users too.

bdk_chain/src/sparse_chain.rs
```diff
@@ -317,60 +298,66 @@ impl<P: ChainPosition> SparseChain<P> {
...
-    ///
-    /// **Warning**: This function modifies the internal state of the chain. You are responsible
-    /// for persisting these changes to disk if you need to restore them.
-    pub fn insert_checkpoint(&mut self, block_id: BlockId) -> Result<bool, InsertCheckpointErr> {
-        if let Some(&old_hash) = self.checkpoints.get(&block_id.height) {
+    /// height with a different hash this will return an error.
+    pub fn insert_checkpoint_changeset(
+        &self,
+        block_id: BlockId,
+    ) -> Result<ChangeSet<P>, UpdateFailure<P>> {
```
LLFourn, line 343: The error here should not change. There is only one thing that can go wrong which is `HashNotMatching`.

bdk_chain/tests/test_chain_graph.rs
```diff
@@ -4,7 +4,7 @@ mod common;
 use bdk_chain::{
     chain_graph::{ChainGraph, ChangeSet, InflateFailure, UnresolvableConflict, UpdateFailure},
     collections::HashSet,
-    sparse_chain,
+    sparse_chain::{self},
```
LLFourn, line 7: unnecessary

bdk_chain/src/keychain/keychain_tracker.rs
```diff
@@ -121,26 +121,30 @@ where
...
+    pub fn insert_tx_changeset(
+        &self,
         tx: Transaction,
-        position: Option<P>,
-    ) -> Result<bool, sparse_chain::InsertTxErr> {
-        let changed = self.chain_graph.insert_tx(tx.clone(), position)?;
-        self.txout_index.scan(&tx);
-        Ok(changed)
+        pos: P,
+    ) -> Result<KeychainChangeSet<K, P>, chain_graph::UpdateFailure<P>> {
```
LLFourn, line 143: The error should not change. `InsertTxErr` correctly captured the things that could go wrong.

bdk_chain/src/chain_graph.rs
```diff
@@ -82,21 +63,42 @@ impl<P: ChainPosition> ChainGraph<P> {
...
+        tx: Transaction,
+        pos: P,
+    ) -> Result<ChangeSet<P>, UpdateFailure<P>> {
+        // only allow displacement of unconfirmed txs
+        let chain_changeset = self.chain.insert_tx_changeset(tx.txid(), pos)?;
+
+        Ok(self
+            .inflate_changeset(chain_changeset, core::iter::once(tx))
+            .map_err(|failure| match failure {
+                InflateFailure::Missing(_) => panic!("no transaction should be missing"),
```
LLFourn, line 82: This should be `unreachable!`

bdk_chain/src/chain_graph.rs
```diff
@@ -82,21 +63,42 @@ impl<P: ChainPosition> ChainGraph<P> {
...
+    /// Determines the changes required to insert a transaction into the inner [`ChainGraph`] and
+    /// [`SparseChain`] at the given `position`.
+    ///
+    /// If inserting it into the chain `position` will result in conflicts, the returned
+    /// [`ChangeSet`] should evict conflicting transactions.
+    pub fn insert_tx_changeset(
+        &self,
+        tx: Transaction,
+        pos: P,
+    ) -> Result<ChangeSet<P>, UpdateFailure<P>> {
```
LLFourn, line 75: The error here is too general. `UpdateFailure` contains a `NotConnected` variant which cannot occur while inserting a tx. This needs to maintain its own error type.

bdk_esplora_example/src/esplora.rs
```diff
@@ -163,16 +168,23 @@ impl Client {
...
-                            match err {
-                                InsertTxErr::TxTooHigh => {
-                                    /* Don't care about new transactions confirmed while syncing */
-                                }
-                                InsertTxErr::TxMoved => {
+                        match update.insert_tx_changeset(tx.to_tx(), confirmation_time) {
+                            Ok(changeset) => update.apply_changeset(changeset).expect("tx changeset should have no missing txs"),
+                            Err(err) => match err {
+                                bdk_chain::chain_graph::UpdateFailure::Chain(chain_err) => match chain_err {
+                                    sparse_chain::UpdateFailure::NotConnected(_) => panic!("NotConnected must not occur as we are only inserting a tx"),
```
LLFourn, line 175: Note the friction created by the erroneous error type.

LLFourn (changes requested):
Thanks. I think having a way of returning a changeset for any kind of change you want to make is going in the right direction. I think we still want to have a way of generating and applying in the same go  to reduce API friction. 

In general it's fine to re-use full "determine changeset" logic in the single insertion methods like `insert_tx_changeset` by creating a single item update where it helps. However, you must map the error type inside the method to exclude errors that cannot happen during that change (use `unreachable!`. Consider that it may not be worth calling a full `determine_changeset` since the error mapping might mean it's simpler just to manually create the changeset.

bdk_esplora_example/src/esplora.rs
```diff
@@ -163,13 +161,19 @@ impl Client {
...
+                                InsertTxFailure::Chain(TxTooHigh { .. }) => {
+                                    /* Chain tip has increased, ignore tx for now */
+                                }
+                                InsertTxFailure::Chain(TxMovedUnexpectedly { .. }) => {
+                                    /* Reorg occured (catch error below), ignore tx for now */
                                 }
-                                InsertTxErr::TxMoved => {
-                                    /* This means there is a reorg, we will catch that below */
+                                InsertTxFailure::UnresolvableConflict(_) => {
+                                    /* Reorg occured (catch error below), ignore tx for now */
```
LLFourn, line 176: I'm not 100% certain this implies there has been a re-org. It just means there is a confirmed transaction that conflicts with it.

evanlinjin: I'm assuming it would either be a reorg, a bug, or a malicious node/server we are connected to, correct? 😅

LLFourn: I think essplora will actually return stuff that has recently been evicted. See #128

LLFourn (approved):
ACK e86659a350c6538186f01e6f31397c7d3c9dae48

PR #117, opened 2022-12-27
Fix derive_next_unused

### LINKED ISSUE #110: Rename derive_next_unused next_unused

rajarshimaitra:
Fixes #110 

I think there was an internal logic problem previously where we were calling `next` twice. Once to check if next is available, and then in the return call in `else` block. This would progress the iterator twice on each call..

Restructured the logic to only call `next` once, and also solve the lifetime problem. This also allows us to remove redundant `clone`s elsewhere.

Renamed `derive_next_unused` to `next_unused`.

### DIFF
--- a/bdk_chain/src/keychain/keychain_txout_index.rs
+++ b/bdk_chain/src/keychain/keychain_txout_index.rs
@@ -245,7 +245,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// ## Panics
     ///
     /// Panics if `keychain` has never been added to the index
-    pub fn derive_next_unused(&mut self, keychain: &K) -> (u32, &Script) {
+    pub fn next_unused(&mut self, keychain: &K) -> (u32, &Script) {
         let need_new = self.keychain_unused(keychain).next().is_none();
         // this rather strange branch is needed because of some lifetime issues
         if need_new {
--- a/bdk_cli_lib/src/lib.rs
+++ b/bdk_cli_lib/src/lib.rs
@@ -175,7 +175,7 @@ where
     let txout_index = &mut keychain_tracker.txout_index;
 
     let new_address = match addr_cmd {
-        AddressCmd::Next => Some(txout_index.derive_next_unused(&Keychain::External)),
+        AddressCmd::Next => Some(txout_index.next_unused(&Keychain::External)),
         AddressCmd::New => Some(txout_index.derive_new(&Keychain::External)),
         _ => None,
     };
@@ -319,9 +319,7 @@ pub fn create_tx<P: ChainPosition>(
     };
 
     let (change_index, change_script) = {
-        let (index, script) = keychain_tracker
-            .txout_index
-            .derive_next_unused(&internal_keychain);
+        let (index, script) = keychain_tracker.txout_index.next_unused(&internal_keychain);
         (index, script.clone())
     };
     let change_plan = bdk_tmp_plan::plan_satisfaction(


evanlinjin:
> I think there was an internal logic problem previously where we were calling next twice. Once to check if next is available, and then in the return call in else block. This would progress the iterator twice on each call..

The `keychain_unused` method is non-mutating, and the iterator returned on each call is different.

> Restructured the logic to only call next once, and also solve the lifetime problem. This also allows us to remove redundant clones elsewhere.

I like this. I would prefer if `derive_new` uses the old `new_spk` (instead of calling `self.inner.script_pubkeys().get(...)`.

Could you rebase your work on top of #116 ? Maybe create a PR onto my fork, thank you.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -237,21 +238,24 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-        let need_new = self.keychain_unused(keychain).next().is_none();
-        // this rather strange branch is needed because of some lifetime issues
-        if need_new {
-            self.derive_new(keychain)
+    pub fn next_unused<'a>(&'a mut self, keychain: &K) -> (u32, Script) {
+        let existing = {
+            self.keychain_unused(keychain)
+                .next()
+                .map(|(i, s)| (i, s.to_owned()))
+        };
```
evanlinjin, line 254: NIT: We don't need these extra brackets.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -3,6 +3,7 @@ use crate::{
     miniscript::{Descriptor, DescriptorPublicKey},
     ForEachTxout, SpkTxOutIndex,
 };
+use alloc::borrow::ToOwned;
```
evanlinjin, line 6: NIT: We don't need to use `ToOwned` as we are just going from `&T` to `T`.

rajarshimaitra: Went ahead with a clone..

bdk_chain/src/keychain/keychain_txout_index.rs
rajarshimaitra: I didn't get it.. Which extra brackets?

rajarshimaitra:
> I like this. I would prefer if derive_new uses the old new_spk (instead of calling self.inner.script_pubkeys().get(...).

Updated.. 

> Could you rebase your work on top of https://github.com/LLFourn/bdk_core_staging/pull/116 ? Maybe create a PR onto my fork, thank you.

Rebased on top of #116

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -231,27 +231,25 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-        let need_new = self.keychain_unused(keychain).next().is_none();
-        // this rather strange branch is needed because of some lifetime issues
-        if need_new {
-            self.derive_new(keychain)
+    pub fn next_unused<'a>(&'a mut self, keychain: &K) -> (u32, Script) {
+        let existing = {
+            self.keychain_unused(keychain)
+                .next()
+                .map(|(i, s)| (i, s.clone()))
+        };
```
evanlinjin, line 248: > I didn't get it.. Which extra brackets?

@rajarshimaitra 

```suggestion
        let existing = self.keychain_unused(keychain)
            .next()
            .map(|(i, s)| (i, s.clone()));
```

rajarshimaitra: Ah right.. I used that to solve that lifetime issue to remove the immutable ref first out of scope,  before taking another mutable ref.. But we don't that anymore as we are cloning data out..

rajarshimaitra: Updated in https://github.com/LLFourn/bdk_core_staging/pull/117/commits/a47909ca82ca7b240b32cdfb9a391a5c3a83d722

LLFourn:
I think we should not clone the spks out here. It's easy for the caller to do this and it indicates that his was not just derived but also stored in the thing itself which is difficult to convey through the name of the method itself. I think maybe you can do it in a less awkward way without double inserting with the `Entry` API?

rajarshimaitra:
> I think we should not clone the spks out here.

The major hurdle I was facing for returning the reference is, that enforces the lifetime within the function scope to be specified such that two different references doesn't fall under same life time. This is a very edgy case of lifetimes and demonstrated in only documented this [rustnomicon post](https://doc.rust-lang.org/nomicon/lifetime-mismatch.html). We cannot take the 2nd mutable reference, without dropping the 1st immutable one, or some how define the lifetime in the scope such that 2 refs don't occur in the same lifetime.. Or else compiler won't allow passing out the ref..

> I think maybe you can do it in a less awkward way without double inserting with the Entry API?

That was my initial approach but it didn't seem to solve the above lifetime problem..

Thus so far "scoping out and copying data" seemed like the only non nasty fix to me.. But ya, this essentially doesn't allow us to pass out the ref any more.. The other option was what you had before,. have double calls and let go of the first call.

LLFourn:
> Another reason I went with copying was we were doing that in most of our call sites anyway, because that's generally is the downstream requirements for many script APIs.

I think the main API we care about here is the `Address` API which takes a reference. I'd be in favor of just doing the rename in this PR.

I'm not sure why @evanlinjin asked you to rebase on top of his draft PR? Can we just drop the changes other than the rename here and merge it.

rajarshimaitra:
@LLFourn reverted back to just renaming..

bdk_cli_lib/src/lib.rs
```diff
@@ -319,9 +319,7 @@ pub fn create_tx<P: ChainPosition>(
     };
 
     let (change_index, change_script) = {
-        let (index, script) = keychain_tracker
-            .txout_index
-            .derive_next_unused(&internal_keychain);
+        let (index, script) = keychain_tracker.txout_index.next_unused(&internal_keychain);
```
LLFourn, line 322: I just realized that we have to persist the keychain index increment here!!!!!
I actually noticed this bug we were missing a change output in my local esplora example db.

I'll try and fix this tomorrow.

LLFourn (approved):
Thanks @rajarshimaitra

evanlinjin:
> I'm not sure why @evanlinjin asked you to rebase on top of his draft PR? Can we just drop the changes other than the rename here and merge it.

The reason is that #116 already takes care of #110 so this is essentially a duplicate effort, except for the changes to reference/no-reference.

LLFourn:
Ok well since #116 has a different focus we'll just merge this now and move on.

PR #118, opened 2022-12-28
Fix add_spk API

### LINKED ISSUE #98: Make SpkTxOutIndex::add_spk handle duplicate spks under different indexes
What should we do here? The simplest thing would be to return a bool or result indicating whether we ignored the request or not because it is already under another index.

rajarshimaitra:
Check if the spk already exists. If yes, returns false, or else add the spk into map and returns true

Fixes #98

### DIFF
--- a/bdk_chain/src/lib.rs
+++ b/bdk_chain/src/lib.rs
@@ -37,7 +37,7 @@ pub mod collections {
     #![allow(dead_code)]
     pub type HashSet<K> = alloc::collections::BTreeSet<K>;
     pub type HashMap<K, V> = alloc::collections::BTreeMap<K, V>;
-    pub use alloc::collections::btree_map::Entry;
+    pub use alloc::collections::btree_map as hash_map;
     pub use alloc::collections::*;
 }
 
@@ -45,7 +45,7 @@ pub mod collections {
 #[cfg(all(feature = "std", not(feature = "hashbrown")))]
 #[doc(hidden)]
 pub mod collections {
-    pub use std::collections::hash_map::Entry;
+    pub use std::collections::hash_map;
     pub use std::collections::*;
 }
 
@@ -57,5 +57,5 @@ pub mod collections {
     pub type HashSet<K> = hashbrown::HashSet<K>;
     pub type HashMap<K, V> = hashbrown::HashMap<K, V>;
     pub use alloc::collections::*;
-    pub use hashbrown::hash_map::Entry;
+    pub use hashbrown::hash_map;
 }
--- a/bdk_chain/src/lib.rs
+++ b/bdk_chain/src/lib.rs
@@ -37,6 +37,7 @@ pub mod collections {
     #![allow(dead_code)]
     pub type HashSet<K> = alloc::collections::BTreeSet<K>;
     pub type HashMap<K, V> = alloc::collections::BTreeMap<K, V>;
+    pub use alloc::collections::btree_map::Entry;
     pub use alloc::collections::*;
 }
 
@@ -44,6 +45,7 @@ pub mod collections {
 #[cfg(all(feature = "std", not(feature = "hashbrown")))]
 #[doc(hidden)]
 pub mod collections {
+    pub use std::collections::hash_map::Entry;
     pub use std::collections::*;
 }
 
@@ -55,4 +57,5 @@ pub mod collections {
     pub type HashSet<K> = hashbrown::HashSet<K>;
     pub type HashMap<K, V> = hashbrown::HashMap<K, V>;
     pub use alloc::collections::*;
+    pub use hashbrown::hash_map::Entry;
 }
--- a/bdk_chain/src/spk_txout_index.rs
+++ b/bdk_chain/src/spk_txout_index.rs
@@ -19,7 +19,7 @@ use bitcoin::{self, OutPoint, Script, Transaction, TxOut, Txid};
 /// chain or unspent etc you must use other sources of information like a [`SparseChain`].
 ///
 /// [`TxOut`]: bitcoin::TxOut
-/// [`add_spk`]: Self::add_spk
+/// [`add_spk`]: Self::insert_script_pubkey
 /// [`Ord`]: core::cmp::Ord
 /// [`scan`]: Self::scan
 /// [`SparseChain`]: crate::sparse_chain::SparseChain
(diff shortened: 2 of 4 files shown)


LLFourn:
Thaks @rajarshimaitra. Shall we also change the name to `insert_script_pubkey` to be consistent with the other apis?

rajarshimaitra:
Done..

bdk_chain/src/lib.rs
```diff
@@ -37,13 +37,15 @@ pub mod collections {
     #![allow(dead_code)]
     pub type HashSet<K> = alloc::collections::BTreeSet<K>;
     pub type HashMap<K, V> = alloc::collections::BTreeMap<K, V>;
+    pub use alloc::collections::btree_map::Entry;
```
evanlinjin, line 40: I think we should remove this.

bdk_chain/src/lib.rs
```diff
@@ -37,13 +37,15 @@ pub mod collections {
...
     pub use alloc::collections::*;
 }
 
 // When we have std use `std`'s all collections
 #[cfg(all(feature = "std", not(feature = "hashbrown")))]
 #[doc(hidden)]
 pub mod collections {
+    pub use std::collections::hash_map::Entry;
```
evanlinjin, line 48: I also think this should be removed.

bdk_chain/src/lib.rs
```diff
@@ -55,4 +57,5 @@ pub mod collections {
     pub type HashSet<K> = hashbrown::HashSet<K>;
     pub type HashMap<K, V> = hashbrown::HashMap<K, V>;
     pub use alloc::collections::*;
+    pub use hashbrown::hash_map::Entry;
```
evanlinjin, line 60:
```suggestion
    pub use hashbrown::hash_map;
```

I think this would make more sense, then the above two lines can be removed and everything will be consistent.

evanlinjin (changes requested):
Concept ACK 6a39119c730bc81ab9395633572302ca56e5f7a0

I like how this is looking.

bdk_chain/src/lib.rs
rajarshimaitra: Not sure why though.. We can, but then we need to specify `which::Entry` at call site and feature gate it.. Not creating feature gated collection imports is what this module was intended to solve.. But maybe I am missing something.. Do you have a better way to call entry without explicit import?

bdk_chain/src/lib.rs
rajarshimaitra: What exactly are you referring as inconsistent?

evanlinjin: I meant that you can remove the two lines above without resulting in inconsistency.

bdk_chain/src/lib.rs
evanlinjin: If someone wants to import both `bdk_core::collections::Entry` (which is only a `btree_map::Entry`) and `std::collections::hash_map::Entry` at the same time, it will get messy. I think having to explicitly state which `Entry` it is, is a good-to-have.

rajarshimaitra: Hmm.. I think I am getting what you are saying.. I tried it out but unfortunately this particular `Entry` itself will have three variants depending on feature list.. And different parent modules are available with different  So there's no way to do it without different feature gating combination at each module we wanna call `Entry` in.. It seems to me this way of consolidating all required imports is better than having various feature gates in different modules.. If later on we want to use both kinda `Entry` together we can explicitly import and give different names to  them here in `collections` module. like `use::std::collections::btree_map::Entry as StdBtreeEntry` or something like that..

bdk_chain/src/lib.rs
rajarshimaitra: Same this as here https://github.com/LLFourn/bdk_core_staging/pull/118?notification_referrer_id=NT_kwDOAi2U5bM1MTkyOTA5MTQ4OjM2NTQxNjY5&notifications_query=reason%3Aparticipating#discussion_r1060037236

bdk_chain/src/spk_txout_index.rs
```diff
@@ -149,13 +146,19 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
...
-        self.spk_indexes.insert(spk.clone(), index.clone());
-        self.script_pubkeys.insert(index.clone(), spk);
-        self.unused.insert(index);
+    pub fn insert_script_pubkey(&mut self, index: I, spk: Script) -> bool {
+        match self.spk_indexes.entry(spk.clone()) {
+            Entry::Vacant(value) => {
+                value.insert(index.clone());
+                self.script_pubkeys.insert(index.clone(), spk);
+                self.unused.insert(index);
+                return true;
```
LLFourn, line 158: no need for explicit return values.

bdk_chain/src/spk_txout_index.rs
```diff
@@ -149,13 +146,19 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
         &self.script_pubkeys
     }
 
-    /// Adds a script pubkey to scan for.
+    /// Adds a script pubkey to scan for. Returns `false` is spk already exists in the map
```
LLFourn, line 149: returns false and **does nothing** (that's the important part!),

bdk_chain/src/lib.rs
LLFourn: We can make `hash_brown` a non-optional dependency if that helps.

LLFourn (commented):
Looks good. Just need to sort out the `Entry` export thing. I think you can just export the `hash_map` module from each thing and import it has collections::hash_map::Entry?

bdk_chain/src/spk_txout_index.rs
rajarshimaitra: Done..

rajarshimaitra:
@LLFourn updated as suggestied. Let me know if this is how you meant to import..  

> We can make hash_brown a non-optional dependency if that helps.

Could be, but what we have currently doesn't seem to be bad either..

LLFourn (approved):
ACK 46524adb19ee61a495bf934293b4f232d1b8020a

evanlinjin (approved):
LGTM!

Issue #85, opened 2022-12-13
Implement updates that returns position in block and merkle proofs


LLFourn:
We have built the internals to allow us to include position in block and even a full merkle proof for the transactoin in `ChainIndex`. Let's do it and show it in an example. 

cc @evanlinjin a decent candidate for this would be creating a `ChainGraph` update from blocks fetched from rpc. It could be done for electrum too (see #77) because iirc getting the position is the same as getting the merkle root. This should be a different call though because you don't always want this.

Another thing worth considering is that including the position in block makes the `ChainIndex` a total order so we may need to extend that trait to tell us whether the index is a total order. This will allow us to make sure no two chain indexes are the same if it is a total order (i.e. no transactions are at the same height in the same block).

PR #119, opened 2023-01-04
demonstrate syncing with `block_position` indexing

rajarshimaitra:
This is an attemted approach at #85 . Not final and can be structured better. Opening up for approach discussions.

Build on top of #77  and demos it over electrum API.

Things Done:
 - new `BlockPosition` struct that is `ChainPosition` with custom ordering.
 - Fetch block position data from electrum and create the index accordingly.
 - New SyncBlockPosition command to sync in this way. This will create a new tracker and a new db. the regular `TxHeight` tracker and db won't be updated in that call.
 
 Question:
  - What are better way to structure this. There's a lot of copy past of existing code, and I couldn't find a good way to satisfy type checker, with the way the main function is structured currently. So just dumped all the code in a separate cli arg.

### DIFF
--- a/bdk_electrum_example/src/electrum.rs
+++ b/bdk_electrum_example/src/electrum.rs
@@ -2,7 +2,7 @@ use std::{collections::BTreeMap, ops::Deref};
 
 use bdk_chain::{
     bitcoin::{BlockHash, Script, Txid},
-    sparse_chain::{ChainPosition, InsertCheckpointFailure, InsertTxFailure, SparseChain},
+    sparse_chain::{self, ChainPosition, SparseChain},
     BlockId, TxHeight,
 };
 use bdk_cli::Broadcast;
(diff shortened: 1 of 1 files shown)


bdk_chain/src/chain_data.rs
```diff
@@ -114,6 +116,96 @@ impl ConfirmationTime {
...
+///
+/// Unconfirmed transactions are treated equal.
+#[derive(Debug, Clone, Eq, Copy, core::hash::Hash)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Deserialize, serde::Serialize),
+    serde(crate = "serde_crate")
+)]
+pub enum BlockPosition {
+    Confirmed { height: u32, position: usize },
```
LLFourn, line 131: why can't this derive ord/partial ord?

rajarshimaitra: Yes it can.. Reverted..

LLFourn:
Can you try and re-do this PR on top of master @rajarshimaitra?

rajarshimaitra:
@LLFourn rebased on master..

bdk_electrum_example/src/electrum.rs
```diff
@@ -182,10 +312,10 @@ impl ElectrumClient {
                     for (txid, pos) in txid_list {
                         if let Err(failure) = sparse_chain.insert_tx(txid, pos) {
                             match failure {
-                                sparse_chain::InsertTxFailure::TxTooHigh { .. } => {
+                                InsertTxFailure::TxTooHigh { .. } => {
                                     unreachable!("We should not encounter this error as we ensured tx_height <= tip.height");
                                 }
-                                sparse_chain::InsertTxFailure::TxMovedUnexpectedly { .. } => {
+                                InsertTxFailure::TxMovedUnexpectedly { .. } => {
```
LLFourn, line 318: nit: I prefer to keep the `sparse_chain::` qualification so I know which module it's from.

PR #121, opened 2023-01-06
Make TxGraph::calculate_fee return 0 for coinbase

LLFourn:
And fix a panic. Arguable this should return a negative value but I've left the API the same for now.

### DIFF
--- a/bdk_chain/tests/test_tx_graph.rs
+++ b/bdk_chain/tests/test_tx_graph.rs
@@ -267,7 +267,7 @@ fn insert_txout_does_not_displace_tx() {
 }
 
 #[test]
-fn test_calcualte_fee() {
+fn test_calculate_fee() {
     let mut graph = TxGraph::default();
     let intx1 = Transaction {
         version: 0x01,
@@ -299,9 +299,9 @@ fn test_calcualte_fee() {
         },
     );
 
-    graph.insert_tx(intx1.clone());
-    graph.insert_tx(intx2.clone());
-    graph.insert_txout(intxout1.0, intxout1.1);
+    let _ = graph.insert_tx(intx1.clone());
+    let _ = graph.insert_tx(intx2.clone());
+    let _ = graph.insert_txout(intxout1.0, intxout1.1);
 
     let mut tx = Transaction {
         version: 0x01,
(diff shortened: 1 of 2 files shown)


bdk_chain/tests/test_tx_graph.rs
```diff
@@ -141,3 +141,97 @@ fn insert_tx_can_retrieve_full_tx_from_graph() {
...
+    assert_eq!(graph.calculate_fee(&tx), Some(100));
+
+    tx.input.remove(2);
+
+    // fee would be negative
+    assert_eq!(graph.calculate_fee(&tx), Some(-200));
+}
+
+#[test]
+fn test_calculate_fee_negative_fee() {}
```
rajarshimaitra, line 220: just to note that this test is already covered in the previous one.

bdk_chain/tests/test_tx_graph.rs
```diff
@@ -141,3 +141,97 @@ fn insert_tx_can_retrieve_full_tx_from_graph() {
...
+
+#[test]
+fn test_calculate_fee_on_coinbase() {
```
rajarshimaitra, line 223: just for the sake of completeness maybe we can also add for the `None` case, where we are missing one/many inputs.

rajarshimaitra (commented):
tACK 60250862631015c2efa9b93a911ad4c1744f4da7

bdk_chain/tests/test_tx_graph.rs
```diff
@@ -141,3 +141,97 @@ fn insert_tx_can_retrieve_full_tx_from_graph() {
     graph.insert_tx(tx.clone());
     assert_eq!(graph.tx(tx.txid()), Some(&tx));
 }
+
+#[test]
+fn test_calcualte_fee() {
```
evanlinjin, line 146:
```suggestion
fn test_calculate_fee() {
```

evanlinjin (approved):
ACK 60250862631015c2efa9b93a911ad4c1744f4da7

Just a NIT and Raj's suggestions on tests

evanlinjin (approved):
ACK 181e0cc4810e56d21d8b7e97f23965fd8096a356

PR #123, opened 2023-01-06
Balance calculations for keychain tracker

LLFourn:
This is ready for review now. Needs the bug fixed by https://github.com/LLFourn/bdk_core_staging/pull/126 to be merged to master first which I think might come with #116 .


bdk_chain/src/keychain.rs
```diff
@@ -91,3 +91,56 @@ impl<K, P> ForEachTxout for KeychainChangeSet<K, P> {
...
+impl Balance {
+    /// Get sum of trusted_pending and confirmed coins
+    pub fn get_spendable(&self) -> u64 {
+        self.confirmed + self.trusted_pending
+    }
+
+    /// Get the whole balance visible to the wallet
+    pub fn get_total(&self) -> u64 {
+        self.confirmed + self.trusted_pending + self.untrusted_pending + self.immature
+    }
```
LLFourn, line 122: These two methods were copy pasted from BDK. I am not sure about this though. Why is `trusted_pending` more "spendable" than `untrusted_pending`? I'd opt to remove `get_spendable` and rename `get_total` to `total`.

cc @afilini

LLFourn: I did the above ^

evanlinjin (approved):
ACK 9671409179d60fb0a20f8a09d816eea202b96131

PR #124, opened 2023-01-09
Test with default features and minimal

LLFourn:
code coverage tests don't do this.

In practice the tests were failing unless --all-features was passed.


LLFourn:
I also made miniscript a default feature

evanlinjin (approved):
ACK bca5ab111d6685e0d39367d7f3b1c07d28222781

Although I couldn't test `.github/workflows/test.yml`...

PR #126, opened 2023-01-09
Chain position refused to change upon inserting it again at a different height

LLFourn:


evanlinjin:
This will conflict with #116 I will cherrypick this commit ontop of #116

evanlinjin:
Added to #166 in commit 526432a18b7ea9540c94e39a35c4c647a44cd74e

PR #134, opened 2023-01-11
Apply conflict rules in ChainGraph::apply_changeset

### LINKED ISSUE #132: chain graph changes should be error free
7440681e88850e438c9e4a938cce021e4714c3ea added an error case for the tx graph struct where it would return an error when a txout "conficts". This seems to mean an update contains a txout not attached a tx that doesn't agree with what was there before.

I don't think we should have this error case. It is more likely to create a security bug from the extra friction it creates than to solve one. Consider that if you already have the full tx and an update contains an erroneous output where as before it would have done the right thing by ignoring it, there is now an error case. If previously there was a dangling txout then we can just replace it with the new one. If the user does not like the ambiguity of these txouts then they can just

LLFourn:
commits:
1. Add test for bug I found where `ChainGraph` would be OK with the `TxGraph` only having txout for the chain transaction rather than a full transaction.
2. Removes error case from TxGraph's API. See #132 
3. Checks for conflicts in ChainGraph::apply_changeset

Note: we currently check for conflicts in the changeset with the chain graph's own state but not within the changeset itself which can lead to an inconsistent ChainGraph (conflicting txs).

Fixes #132

### DIFF
--- a/bdk_chain/src/chain_graph.rs
+++ b/bdk_chain/src/chain_graph.rs
@@ -4,6 +4,7 @@ use crate::{
     tx_graph::{self, TxGraph},
     BlockId, ForEachTxout, FullTxOut, TxHeight,
 };
+use alloc::{string::ToString, vec::Vec};
 use bitcoin::{OutPoint, Transaction, TxOut, Txid};
 use core::fmt::Debug;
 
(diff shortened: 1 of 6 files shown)


evanlinjin:
@LLFourn What was the conclusion of our talk this morning? Are we planning to include or exclude all errors, or have the ability to check `Additions` for any non-monotone changes?

As non-monotone behavior of `TxGraph` will result in non-monotonicity of structures such as `TxOutIndex` and `ChainGraph` (that depend on `TxGraph`).

LLFourn:
@evanlinjin Since we model the blockchain data source as always giving correct info in general I think we should just not have error cases when the blockchain contradicts itself for now. I think the right way to solve this for users who want guarantees that this can't happen is to use full transactions only.

evanlinjin (approved):
ACK 37ae14b874240ab7b09038ebe74351561ac68bf2

LLFourn:
In the end @evanlinjin and I decided to remove this conflict resolution stuff in apply changeset and follow the approach that "apply changeset" is only safe if you got the changeset from the creation method itself. This keeps this simpler for now.

Issue #120, opened 2023-01-04
Super-issue regarding transaction history consistency


evanlinjin:
As mentioned by @LLFourn in a set of messages, there are scenarios which we should handle:

- [x] When we evict a tx due to conflict, we should also evict the tx's children.
- [x] When we insert a tx without chain position that exposes 2 `TxHeight::Unconfirmed` txs have conflict, we should have a way to ensure consistency.

Also related:

- [ ] We should implement CPFP feerate logic.

### Contextual quotes

> This is why I am concerned about just allowing arbitrary graph insertions.
> We have to come up with an algorithm to always make txs in chain consistent with graph on every insertion.
> -- @LLFourn

LLFourn:
To give further context. Inserting a tx into a graph without a position can make two transactions that are currently non-conflicting conflict if that new tx is the ancestor of one tx and conflicts with the ancestor of another. So if we want to allow positionless graph insertions we must not allow it directly into the graph. There is also an error case when both conflicting transactions are confirmed already.

So anytime we insert a tx at chain position we need to evict conflicts and all their children in the chain. However we need to walk all the child nodes in the DAG in the `TxGraph` to find the transactions to evict (note we don't care if the all children are in the chain or not -- perhaps a child is not in the chain but a grandchild is, we still need to evict the grandchild).

If we allow positionless insertion into the graph we still might need to do eviction. If the tx you are inserting into the graph does conflict with a tx that is in the chain then you need to evict all child transactions of that new graph transaction from the chain before inserting it into the graph.

Furthermore, any time you insert a new tx into the chain which spends from an output that is not in the chain but is in the graph you need to check whether the graph allows its parent transactions to be in the chain or not. If the parent couldln't be in the chain then that new tx can't be in their either. I think this check needs to be done regardless of whether we allow permissionless insertion.

Possible I'm missing something but that might be it.

LLFourn:
Another problem: https://github.com/LLFourn/bdk_core_staging/issues/135

evanlinjin:
Completed by #138, the last missing piece is CPFP feerate calculation logic. Will make a separate ticket for this.

LLFourn:
I don't think we can say this is really dealt with. Do we have a test case for:

[...]

PR #138, opened 2023-01-12
`TxGraph` descendants

evanlinjin:
This PR currently implements:
* Evicting descendants of conflicting transactions (before, we only evicted direct conflicts). -- #120
* The evict-descendant logic is now used to fix: #139
* Traversing descendants of given `Txid`.

### DIFF
--- a/bdk_chain/src/chain_graph.rs
+++ b/bdk_chain/src/chain_graph.rs
@@ -174,16 +174,19 @@ impl<P: ChainPosition> ChainGraph<P> {
     }
 
     /// Given a transaction, return an iterator of in-chain [`Txid`]s that conflict with it (spends
-    /// at least one of the same inputs).
+    /// at least one of the same inputs). This includes all descendants of conflicting transactions.
     ///
-    /// This method is comparable to [`TxGraph::conflicting_txids()`] which returns all conflicting
-    /// transactions, whereas this method only returns conflicts that exist in the [`SparseChain`].
-    pub fn conflicting_txids_in_chain<'a>(
+    /// This method is comparable to [`TxGraph::tx_conflicts()`] which returns all conflicting
+    /// transactions (including descendants), whereas this method only returns conflicts that exist
+    /// in the [`SparseChain`].
+    pub fn tx_conflicts<'a>(
         &'a self,
         tx: &'a Transaction,
     ) -> impl Iterator<Item = (&'a P, Txid)> + 'a {
         self.graph
-            .conflicting_txids(tx)
+            // Using `tx_graph::no_filter` is less efficient than filtering out txs that aren't in
+            // chain. However, it can potentially be more reliable.
+            .tx_conflicts(tx, tx_graph::no_filter)
             .filter_map(|(_, conflicting_txid)| {
                 self.chain
                     .tx_position(conflicting_txid)
@@ -216,7 +219,7 @@ impl<P: ChainPosition> ChainGraph<P> {
                 full_tx
             })
             .flat_map(|(new_txid, new_tx, new_pos)| {
-                self.conflicting_txids_in_chain(new_tx)
+                self.tx_conflicts(new_tx)
                     .map(move |(conflict_pos, conflict_txid)| {
                         (new_pos.clone(), new_txid, conflict_pos, conflict_txid)
                     })
--- a/bdk_chain/src/chain_graph.rs
+++ b/bdk_chain/src/chain_graph.rs
@@ -173,25 +173,20 @@ impl<P: ChainPosition> ChainGraph<P> {
         Ok(changeset)
     }
 
-    /// Given a transaction, return an iterator of in-chain [`Txid`]s that conflict with it (spends
-    /// at least one of the same inputs). This includes all descendants of conflicting transactions.
+    /// Given a transaction, return an iterator of `txid`s that conflict with it (spends at least
+    /// one of the same inputs). This includes all descendants of conflicting transactions.
     ///
-    /// This method is comparable to [`TxGraph::tx_conflicts()`] which returns all conflicting
-    /// transactions (including descendants), whereas this method only returns conflicts that exist
-    /// in the [`SparseChain`].
-    pub fn tx_conflicts<'a>(
+    /// This method only returns conflicts that exist in the [`SparseChain`] as transactions that
+    /// are not included in [`SparseChain`] are already considered as evicted.
+    pub fn tx_conflicts_in_chain<'a>(
         &'a self,
         tx: &'a Transaction,
     ) -> impl Iterator<Item = (&'a P, Txid)> + 'a {
-        self.graph
-            // Using `tx_graph::no_filter` is less efficient than filtering out txs that aren't in
-            // chain. However, it can potentially be more reliable.
-            .tx_conflicts(tx, tx_graph::no_filter)
-            .filter_map(|(_, conflicting_txid)| {
-                self.chain
-                    .tx_position(conflicting_txid)
-                    .map(|conflicting_pos| (conflicting_pos, conflicting_txid))
-            })
+        self.graph.walk_conflicts(tx, |_, conflict_txid| {
+            self.chain
+                .tx_position(conflict_txid)
+                .map(|conflict_pos| (conflict_pos, conflict_txid))
+        })
     }
 
     /// Fix changeset conflicts.
@@ -219,7 +214,7 @@ impl<P: ChainPosition> ChainGraph<P> {
                 full_tx
             })
             .flat_map(|(new_txid, new_tx, new_pos)| {
-                self.tx_conflicts(new_tx)
+                self.tx_conflicts_in_chain(new_tx)
                     .map(move |(conflict_pos, conflict_txid)| {
                         (new_pos.clone(), new_txid, conflict_pos, conflict_txid)
                     })
@@ -501,7 +496,7 @@ impl<P: core::fmt::Debug> core::fmt::Display for UnresolvableConflict<P> {
             already_confirmed_tx,
             update_tx,
         } = self;
-        write!(f, "update transaction {} at height {:?} conflicts with an already confirmed transaction {} at height {:?}", 
+        write!(f, "update transaction {} at height {:?} conflicts with an already confirmed transaction {} at height {:?}",
             update_tx.1, update_tx.0, already_confirmed_tx.1, already_confirmed_tx.0)
     }
 }
--- a/bdk_chain/src/tx_graph.rs
+++ b/bdk_chain/src/tx_graph.rs
@@ -1,11 +1,17 @@
+use core::ops::RangeInclusive;
+
 use crate::{collections::*, ForEachTxout};
-use alloc::{borrow::Cow, vec::Vec};
-use bitcoin::{OutPoint, Transaction, TxIn, TxOut, Txid};
+use alloc::vec::Vec;
+use bitcoin::{OutPoint, Transaction, TxOut, Txid};
 
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct TxGraph {
     txs: HashMap<Txid, TxNode>,
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
+
+    // This atrocity exists so that `TxGraph::outspends()` can return a reference.
+    // FIXME: This can be removed once `HashSet::new` is a const fn.
+    empty_outspends: HashSet<Txid>,
 }
 
 /// Node of a [`TxGraph`]
@@ -21,29 +27,13 @@ impl Default for TxNode {
     }
 }
 
-/// Do not filter any descendants.
-///
-/// This is a convenience method intended for use with [`TxGraph::iter_tx_descendants()`] and
-/// related methods.
-pub fn no_filter(_: usize, _: Txid) -> bool {
-    true
-}
-
 impl TxGraph {
     /// The transactions spending from this output.
     ///
     /// `TxGraph` allows conflicting transactions within the graph. Obviously the transactions in
     /// the returned will never be in the same blockchain.
-    ///
-    /// Note this returns a [`Cow`] because of an implementation detail.
-    ///
-    /// [`Cow`]: std::borrow::Cow
-    // FIXME: this Cow could be gotten rid of if we could do HashSet::new in a const fn
-    pub fn outspends(&self, outpoint: OutPoint) -> Cow<HashSet<Txid>> {
-        self.spends
-            .get(&outpoint)
-            .map(|outspends| Cow::Borrowed(outspends))
-            .unwrap_or(Cow::Owned(HashSet::default()))
+    pub fn outspends(&self, outpoint: OutPoint) -> &HashSet<Txid> {
+        self.spends.get(&outpoint).unwrap_or(&self.empty_outspends)
     }
 
     /// The transactions spending from `txid`.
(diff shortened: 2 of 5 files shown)


LLFourn:
So the point of this is to fix some things in #120 like evicting children of an evicted tx right. Cool. Before merging we should make the algorithm non-recursive because rust doesn't support tail call optimization yet I think.

rajarshimaitra (commented):
Concept ACK.. I have yet to go through the tests.. Overall it looks good and is a much-required tool.. 

Below are some conceptual questions..

bdk_chain/src/tx_graph.rs
```diff
@@ -348,3 +402,52 @@ impl ForEachTxout for Additions {
...
+                    .map(|&txid| (depth + 1, txid))
+                    .collect::<Vec<_>>();
+                self.stack.append(&mut children);
+                return Some((depth, txid));
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.stack.len(), None)
```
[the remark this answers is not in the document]
evanlinjin: This is part of the `Iterator` trait. Please refer to docs: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

Whether my implementation here is correct or not is another question, I'm realizing now that my implementation may be incorrect.

bdk_chain/src/tx_graph.rs
```diff
@@ -348,3 +402,52 @@ impl ForEachTxout for Additions {
...
+            path_filter,
+            visited: [].into(),
+            stack: [(0, txid)].into(),
+        }
+    }
+}
+
+impl<'a, F> Iterator for TxDescendants<'a, F>
+where
+    F: FnMut(usize, Txid) -> bool,
```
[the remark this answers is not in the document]
evanlinjin: My intention is that no one can construct `TxDescendants` without calling methods from `TxGraph`. From my understanding, this is idiomatic.

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +175,84 @@ impl TxGraph {
...
-    pub fn conflicting_txids<'g>(
+    /// Iterates descendant transactions of given `txid` (including the given `txid`).
+    ///
+    /// if `path_filter` returns `true`, we will continue exploring the path from the `txid` of
+    /// `depth`, where `depth: usize` is the distance from the initial `txid` to the descendant.
+    pub fn iter_tx_descendants<'g>(
+        &'g self,
+        txid: Txid,
+        path_filter: impl FnMut(usize, Txid) -> bool + 'g,
+    ) -> impl Iterator<Item = (usize, Txid)> + '_ {
```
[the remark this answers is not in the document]
evanlinjin: Whether we know the full descendant graph or not is not relevant. `TxGraph` will contain conflicting spends within itself. The idea of `path_filter` is we can start from a `txid` and ignore paths that are irrelevant to us.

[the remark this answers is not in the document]
evanlinjin: Currently `path_filter` maybe sub-optimal, because we filter paths individually, instead of choosing between "conflicting paths", which can help us construct consistent histories.

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +175,84 @@ impl TxGraph {
...
+
+    /// Iterates over all conflicting txids (including descendants) of the given transaction.
+    ///
+    /// The `path_filter` filters descendant paths to check. Refer to
+    /// [`Self::iter_tx_descendants()`] for more details.
+    pub fn tx_conflicts<'g>(
         &'g self,
         tx: &'g Transaction,
+        path_filter: impl Fn(usize, Txid) -> bool + 'g,
     ) -> impl Iterator<Item = (usize, Txid)> + '_ {
```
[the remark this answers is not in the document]
evanlinjin: The first field of the tuple is the `vin` of the input `tx`. We will need to document this clearer.

bdk_chain/src/tx_graph.rs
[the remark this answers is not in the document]
rajarshimaitra: Okay.. It's still not fully clear to me how a user can ever provide a path filter without analyzing the full graph himself or by making some other conflict identification code with the full graph data.. But looks okay for now.. I suppose in most part of the code we are gonna use `no_filter` anyway..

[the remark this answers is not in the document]
evanlinjin: > It's still not fully clear to me how a user can ever provide a path filter without analyzing the full graph himself 

One way is using the sparsechain.

I agree that the implementation is sub-optimal. Open for ideas.

bdk_chain/src/tx_graph.rs
```diff
@@ -1,11 +1,17 @@
...
+use alloc::vec::Vec;
+use bitcoin::{OutPoint, Transaction, TxIn, TxOut, Txid};
 
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct TxGraph {
     txs: HashMap<Txid, TxNode>,
     spends: BTreeMap<OutPoint, HashSet<Txid>>,
+
+    // This atrocity exists so that `TxGraph::outspends()` can return a reference.
+    // FIXME: This can be removed once we are able to have a static `HashSet` (somehow).
```
LLFourn, line 13: "once HashSet::new is a const fn"

bdk_chain/src/tx_graph.rs
```diff
@@ -22,20 +28,17 @@ impl Default for TxNode {
 }
 
 impl TxGraph {
+    /// The total number of transactions (full and partial) that exists in the [`TxGraph`].
+    pub fn count(&self) -> usize {
+        self.txs.len()
+    }
```
LLFourn, line 34: Why do we need this?

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +170,87 @@ impl TxGraph {
...
         self.txs.iter().filter_map(|(txid, tx)| match tx {
             TxNode::Whole(_) => None,
             TxNode::Partial(partial) => Some((*txid, partial)),
         })
     }
 
-    /// Return an iterator of conflicting txids, where the first field of the tuple is the vin of
-    /// the original tx in which the txid conflicts.
-    pub fn conflicting_txids<'g>(
+    /// Creates an iterator that both filters and maps descendants of the given `txid`.
```
LLFourn, line 189: When I look at the closure it gives me `OutPointSpends` but from the docs I am expecting a `Transaction` given that `Transaction`s are the "descendants".

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +170,87 @@ impl TxGraph {
...
+
+        TxDescendants::from_spends(self, spends, [tx.txid()].into(), filter_map)
+    }
+
+    /// Given a `txin` and the `txid` of the transaction in which it resides, return an iterator of
+    /// txids which directly conflict with the given `txin`.
+    ///
+    /// **WARNING:** It is up to the caller to ensure the given `txin` actually resides in `txid`.
+    /// **WARNING:** This only returns directly conflicting txids and does not include descendants
+    /// of those txids.
```
LLFourn, line 225: These warnings make me think this API shouldn't exist. Why is this useful? Also why do I have to provide a `TxIn` rather than an `OutPoint`.

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +170,87 @@ impl TxGraph {
...
+            .cloned()
+    }
+
+    /// Given a transaction, return an iterator of txids which directly conflict with the given
+    /// transaction's inputs (spends). The conflicting txids are returned with the given
+    /// transaction's vin (in which it conflicts).
+    ///
+    /// **WARNING:** This only returns directly conflicting txids and does not include descendants
+    /// of those txids.
+    pub fn adjacent_conflicts_of_tx<'g>(
```
LLFourn, line 245:
```suggestion
    pub fn direct_conflicts_of_tx<'g>(
```
You use the term direct in the doc comment to describe this. I think you should call it that.

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +170,87 @@ impl TxGraph {
...
+            .flatten()
+            .filter(move |&&conflicting_txid| conflicting_txid != txid)
+            .cloned()
+    }
+
+    /// Given a transaction, return an iterator of txids which directly conflict with the given
+    /// transaction's inputs (spends). The conflicting txids are returned with the given
+    /// transaction's vin (in which it conflicts).
+    ///
+    /// **WARNING:** This only returns directly conflicting txids and does not include descendants
```
LLFourn, line 243: The `**WARNING**` seems a bit much.

```suggestion
    /// This only returns directly conflicting txids and does not include descendants of those transactions.
```

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +170,87 @@ impl TxGraph {
...
-    /// the original tx in which the txid conflicts.
-    pub fn conflicting_txids<'g>(
+    /// Creates an iterator that both filters and maps descendants of the given `txid`.
+    ///
+    /// When the supplied closure returns `None`, spend paths from the given descendant are not
+    /// explored, and the descendant will not be yielded by the iterator.
+    pub fn filter_map_descendants<'g, T>(
+        &'g self,
+        txid: Txid,
+        filter_map: impl FnMut(OutPointSpends, Txid) -> Option<T> + 'g,
```
LLFourn, line 196: Why does the closure need to be a filter map? Can't it just do filtering.

I think what would make sense to me is:

```
fn walk_descendants(&self, txid: Txid, should_walk: FnMut(&Transaction, u32) -> bool) -> impl Iterator<Item=&Transaction> {
...
}
```

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +170,87 @@ impl TxGraph {
...
+    ) -> impl Iterator<Item = T> + '_ {
```
LLFourn, line 197: Instead of returning an `impl Iterator` you should return `TxDescendants` explicitly.

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +170,87 @@ impl TxGraph {
...
+        TxDescendants::new(self, txid, filter_map)
+    }
+
+    /// Creates an iterator that both filters and maps conflicting transactions (this includes
+    /// descendants of adjacently-conflicting transactions).
+    ///
+    /// When the supplied closure returns `None`, spend paths from the given descendant are not
+    /// explored, and the descendant will not be yielded by the iterator.
+    pub fn filter_map_conflicts<'g, T: 'g>(
```
LLFourn, line 206: Similar comment. Just allowing walking the conflicts.

LLFourn (changes requested):
Nice idea to turn conflicts/descendants into an iterator. I've left comments on we can more fully realize that idea.

bdk_chain/src/tx_graph.rs
LLFourn: Maybe we can just not have a size hint for now.

The correct solution to this is to store full txs and outputs in different maps so you can easily count how many you have of each. These seems out of scope for this PR so I'd remove this and the size hint.

bdk_chain/src/tx_graph.rs
evanlinjin: I think it is more helpful and flexible as a filter map. For example, when `ChainGraph` detects conflicts of a given transaction, we only want to include conflicts that are in `SparseChain` (filter). Additionally, we want the conflict to be returned with the `ChainPosition` (map).

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +165,70 @@ impl TxGraph {
...
+    /// Creates an iterator that both filters and maps descendants from the starting `txid`.
+    ///
+    /// The supplied closure takes in two inputs `(depth, descendant_txid)`:
+    ///
+    /// * `depth` is the distance between the starting `txid` and the `descendant_txid`. I.e. if the
+    ///     descendant is spending an output of the starting `txid`, the `depth` will be 1.
+    /// * `descendant_txid` is the descendant's txid which we are considering to walk.
+    ///
+    /// The supplied closure returns an `Option<T>`, allowing the caller to specify how the
+    /// descendants returned by the iterator will be represented. When the supplied closure returns
```
LLFourn, line 193:
```suggestion
    /// The supplied closure returns an `Option<T>`, allowing the caller to map each node it visits and decide whether to visit descendants. 
```

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +165,70 @@ impl TxGraph {
...
+    /// `None`, spend paths from the given descendant will not be explored, and the descendant will
+    /// not be yielded by the iterator.
+    pub fn iter_descendants<'g, F, T>(&'g self, txid: Txid, walk_map: F) -> TxDescendants<F>
+    where
+        F: FnMut(usize, Txid) -> Option<T> + 'g,
```
LLFourn, line 198: Why can't this just be a `&Transaction`?

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +165,70 @@ impl TxGraph {
...
+    ///
+    /// * `depth` is the distance between the starting `txid` and the `descendant_txid`. I.e. if the
+    ///     descendant is spending an output of the starting `txid`, the `depth` will be 1.
+    /// * `descendant_txid` is the descendant's txid which we are considering to walk.
+    ///
+    /// The supplied closure returns an `Option<T>`, allowing the caller to specify how the
+    /// descendants returned by the iterator will be represented. When the supplied closure returns
+    /// `None`, spend paths from the given descendant will not be explored, and the descendant will
+    /// not be yielded by the iterator.
+    pub fn iter_descendants<'g, F, T>(&'g self, txid: Txid, walk_map: F) -> TxDescendants<F>
```
LLFourn, line 196:
```suggestion
    pub fn walk_descendants<'g, F, T>(&'g self, txid: Txid, walk_map: F) -> TxDescendants<F>
```

bdk_chain/src/tx_graph.rs
```diff
@@ -167,38 +165,70 @@ impl TxGraph {
...
+        F: FnMut(usize, Txid) -> Option<T> + 'g,
+    {
+        TxDescendants::new_exclude_root(self, txid, walk_map)
+    }
+
+    /// Creates an iterator that both filters and maps conflicting transactions (this includes
+    /// descendants of directly-conflicting transactions, which are also considered conflicts).
+    ///
+    /// Refer to [`Self::iter_descendants()`] for `walk_map` usage.
+    pub fn iter_conflicts<'g, F, T>(&'g self, tx: &'g Transaction, walk_map: F) -> TxDescendants<F>
```
LLFourn, line 207:
```suggestion
    pub fn walk_conflicts<'g, F, T>(&'g self, tx: &'g Transaction, walk_map: F) -> TxDescendants<F>
```

bdk_chain/src/tx_graph.rs
```diff
@@ -348,3 +378,124 @@ impl ForEachTxout for Additions {
...
+    Some(txid)
+}
+
+/// Walk all descendants and include the `depth`.
+///
+/// This is a convenience method intended for use with [`TxGraph::iter_descendants()`] and
+/// related methods.
+pub fn walk_all_with_depth(depth: usize, txid: Txid) -> Option<(usize, Txid)> {
+    Some((depth, txid))
+}
```
LLFourn, line 497: I think we can delete these. Not worth the documentation surface area.

bdk_chain/src/chain_graph.rs
```diff
@@ -173,22 +173,20 @@ impl<P: ChainPosition> ChainGraph<P> {
...
-    /// at least one of the same inputs).
+    /// Given a transaction, return an iterator of `txid`s that conflict with it (spends at least
+    /// one of the same inputs). This includes all descendants of conflicting transactions.
     ///
-    /// This method is comparable to [`TxGraph::conflicting_txids()`] which returns all conflicting
-    /// transactions, whereas this method only returns conflicts that exist in the [`SparseChain`].
-    pub fn conflicting_txids_in_chain<'a>(
+    /// This method only returns conflicts that exist in the [`SparseChain`] as transactions that
+    /// are not included in [`SparseChain`] are already considered as evicted.
+    pub fn tx_conflicts_in_chain<'a>(
```
LLFourn, line 181: We should have a test for this that evicts descendants.

LLFourn (approved):
LGTM. Needs test for new chain graph method added. Merge when you are happy @evanlinjin.

bdk_chain/src/tx_graph.rs
evanlinjin: Addressed in DMs. This will be added cost due to lookup.

bdk_chain/src/chain_graph.rs
evanlinjin: https://github.com/LLFourn/bdk_core_staging/blob/842cfc8b50e1eca5b1d83b598eb3d2e4ae2147ca/bdk_chain/tests/test_chain_graph.rs#L350-L354

We currently do

bdk_chain/src/tx_graph.rs
evanlinjin: Done

bdk_chain/src/tx_graph.rs
evanlinjin: Done

bdk_chain/src/tx_graph.rs
evanlinjin: Done

bdk_chain/src/tx_graph.rs
evanlinjin: Resolved in DMs, @LLFourn agrees with me.

bdk_chain/src/tx_graph.rs
evanlinjin: Resolved in DMs.

PR #142, opened 2023-01-13
Improve file store ergonomics

LLFourn:
We need to be able to open a file without loading it yet.

Added is_empty methods on some types that I didn't end up using but kept them because it seemed helpful.

Removed the caching of derivation indices within the store because it wasn't really doing it's job with the cli example and we should do this properly on the outside.

### DIFF
--- a/bdk_chain/src/file_store.rs
+++ b/bdk_chain/src/file_store.rs
@@ -3,18 +3,18 @@ use crate::{
     keychain::{KeychainChangeSet, KeychainTracker},
     sparse_chain,
 };
+use core::marker::PhantomData;
 use std::{
     fs::{File, OpenOptions},
     io::{self, Seek},
     path::Path,
 };
 
-/// Persists changes made to a [`KeychainTracker`] to a file so they can be restored later on.
+/// Persists an append only list of `KeychainChangeSet<K,P>` to a single file.
+/// [`KeychainChangeSet<K,P>`] record the changes made to a [`KeychainTracker<K,P>`].
 #[derive(Debug)]
 pub struct KeychainStore<K, P> {
     db_file: File,
-    /// A cache of what the current state of the derivation indexes are on-disk
-    deriviation_index_cache: BTreeMap<K, u32>,
     chain_index: core::marker::PhantomData<(K, P)>,
 }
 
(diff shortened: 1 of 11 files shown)


LLFourn:
I just rebased on master and cleaned up some stuff from #132. I removed the missing check from apply changeset since we agreed that we should either check all aspects of validity before applying a changeset or none of them.

bdk_chain/src/file_store.rs
```diff
@@ -24,52 +24,85 @@ where
...
-            chain_index: Default::default(),
-        })
+        (changeset, result)
     }
 
+    /// Reads and applies all the changesets stored sequentially to tracker, stopping when it fails
+    /// to read the next one.
+    ///
+    /// **WARNING**: This method changes the write position of the underlying file. The next
+    /// changeset will be written over the erroring entry (or the end of the file if none existed).
```
evanlinjin, line 94: The dangers of doing this without truncating invalid `ChangeSet`s may result in loading an invalid tracker.

Imagine the following scenario:

* Changeset 21 is invalid data (but changeset 22 is valid), so we load tracker from changesets 0 to 20.
* The applications appends a changeset (21b) that is exactly the same size as (21).
* Application closes.
* Application reopens and loads tracker with changesets 0 to 22 (inclusive of 21b). It is possible at this point, that loading 21b and 22 in sequence results in inconsistent tracker.

However, we can't just truncate automatically when there is invalid data as that is dangerous. I think the best way to do this is to have a third field in `KeychainStore` that records whether there has been an error and at which file position. We add another method that can truncate problematic portions of the file (that also clears the error field). If the error field is some, many methods will fail.

evanlinjin (commented):
I have some concerns, although I like the changes overall.

bdk_chain/src/file_store.rs
evanlinjin: As discussed with @LLFourn, instead of adding a field error (in case of opening up a wrong file), we will introduce some magic bytes to the start of the file, thus reducing the risk of accidental file truncation.

bdk_chain/src/file_store.rs
```diff
@@ -24,52 +24,85 @@ where
...
+    /// [`File`]: std::fs::File
+    pub fn new(file: File) -> Self {
+        Self {
+            db_file: file,
+            chain_index: Default::default(),
+        }
+    }
+
+    /// Creates or loads a a store from `db_path`. If no file exists there it will be created.
+    pub fn new_from_path(db_path: &Path) -> Result<Self, io::Error> {
```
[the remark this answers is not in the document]
evanlinjin: This is my misunderstanding of `Path`. However, the `.clone()` below is unnecessary.

evanlinjin (approved):
ACK ded648db2d5dcfd216eaf898764c129b4c5e1684

We will do https://github.com/LLFourn/bdk_core_staging/pull/142#discussion_r1070244344 in a separate PR.

PR #144, opened 2023-01-13
test: Add case to cover unknown outpoint in `calculate_fee`

### LINKED ISSUE #140: Add `TxGraph` fee test with missing inputs
just for the sake of completeness maybe we can also add for the `None` case, where we are missing one/many inputs.

_Originally posted by @rajarshimaitra in https://github.com/LLFourn/bdk_core_staging/pull/121#discussion_r1066897962_

rajarshimaitra:
If we have an unknown transaction input, i.e an outpoint we don't have the `TxOut` for, `calculate_fee` should return `None`. 

Fixes #140 .


evanlinjin (approved):
ACK e9b2d7f367ad110c3258ebdeba0907484ae6e6ed

Thanks @rajarshimaitra !

PR #147, opened 2023-01-14
Add `invalidate_checkpoints..` methods to `ChainGraph`

### LINKED ISSUE #129: Add ChainGraph::invalidate_checkpoints
We added this to `SparseChain` but not chain graph but I think we need it directly on chain graph as well because we can't get a mutable ref to `SparseChain` from the chain graph.

giosue21m:
Closes #129


evanlinjin (approved):
ACK d05d0b235bf111719420b4e01612879766c507e6

PR #149, opened 2023-01-15
Truncate after `KeychainStore::append_changeset` and magic bytes

giosue21m:
Without truncating after appending a changeset, we may result in a sequence of changesets that loads an inconsistent tracker.

Magic bytes are introduced to avoid accidentally truncating non-bdk_file_store files.

### DIFF
--- a/bdk_chain/src/file_store.rs
+++ b/bdk_chain/src/file_store.rs
@@ -124,7 +124,10 @@ where
         Ok(())
     }
 
-    /// Append a new changeset to the file.
+    /// Append a new changeset to the file and truncate file to the end of the appended changeset.
+    ///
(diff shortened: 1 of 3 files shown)


bdk_chain/src/file_store.rs
```diff
@@ -107,22 +129,30 @@ where
         &mut self,
         changeset: &KeychainChangeSet<K, P>,
     ) -> Result<(), io::Error> {
-        if !changeset.is_empty() {
```
LLFourn, line 110: ^ might wanna expand documentation to explain that this function will always set the length of the file to end after the changeset you added.

LLFourn (approved):
Great work.

PR #150, opened 2023-01-15
Methods that change derivation indicies should return changesets

### LINKED ISSUE #141: new address and next_unused should return a changeset
We **need** to remember these changes and apply them to disk.

evanlinjin:
Fixes #141

Main changes:
* Introduce `keychain::DerivationAdditions`. This can be used a indicies changeset or the actual indicies itself. The intention is to have a dedicated structure for the indicies changeset.
* `KeychainTxOutIndex` methods that mutate derivation indices now returns `DerivationAdditions`.

Bug fixes:
* Some `KeychainTxOutIndex` methods did not consider descriptors wildcards. This may lead to unexpected panic.

Other changes:
* Remove `KeychainStore::set_derivation_indices` as `append_changeset achieves the same goal.

### DIFF
--- a/bdk_chain/src/keychain.rs
+++ b/bdk_chain/src/keychain.rs
@@ -32,15 +32,16 @@ pub use keychain_txout_index::*;
         )
     )
 )]
-pub struct DerivationIndices<K>(BTreeMap<K, u32>);
+#[must_use]
+pub struct DerivationAdditions<K>(BTreeMap<K, u32>);
 
-impl<K> DerivationIndices<K> {
+impl<K> DerivationAdditions<K> {
     pub fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
 }
 
-impl<K: Ord> DerivationIndices<K> {
+impl<K: Ord> DerivationAdditions<K> {
     pub fn append(&mut self, mut other: Self) {
         self.0.iter_mut().for_each(|(key, index)| {
             if let Some(other_index) = other.0.remove(key) {
@@ -52,25 +53,34 @@ impl<K: Ord> DerivationIndices<K> {
     }
 }
 
-impl<K> Default for DerivationIndices<K> {
+impl<K> Default for DerivationAdditions<K> {
     fn default() -> Self {
         Self(Default::default())
     }
 }
 
-impl<K> From<BTreeMap<K, u32>> for DerivationIndices<K> {
-    fn from(value: BTreeMap<K, u32>) -> Self {
-        Self(value)
+// impl<K> From<BTreeMap<K, u32>> for DerivationAdditions<K> {
+//     fn from(value: BTreeMap<K, u32>) -> Self {
+//         Self(value)
+//     }
+// }
+
+impl<K: Ord, I> From<I> for DerivationAdditions<K>
+where
+    I: IntoIterator<Item = (K, u32)>,
+{
+    fn from(value: I) -> Self {
+        Self(value.into_iter().collect())
     }
 }
 
-impl<K> AsRef<BTreeMap<K, u32>> for DerivationIndices<K> {
+impl<K> AsRef<BTreeMap<K, u32>> for DerivationAdditions<K> {
     fn as_ref(&self) -> &BTreeMap<K, u32> {
         &self.0
     }
 }
 
-impl<K> AsMut<BTreeMap<K, u32>> for DerivationIndices<K> {
+impl<K> AsMut<BTreeMap<K, u32>> for DerivationAdditions<K> {
     fn as_mut(&mut self) -> &mut BTreeMap<K, u32> {
         &mut self.0
     }
@@ -82,7 +92,7 @@ pub struct KeychainScan<K, P> {
     /// The update data in the form of a chain that could be applied
     pub update: ChainGraph<P>,
     /// The last active indexes of each keychain
-    pub last_active_indexes: DerivationIndices<K>,
+    pub last_active_indexes: DerivationAdditions<K>,
 }
 
 impl<K, I> Default for KeychainScan<K, I> {
@@ -109,7 +119,7 @@ impl<K, I> Default for KeychainScan<K, I> {
 #[must_use]
 pub struct KeychainChangeSet<K, P> {
     /// The changes in local keychain derivation indices
-    pub derivation_indices: DerivationIndices<K>,
+    pub derivation_indices: DerivationAdditions<K>,
     /// The changes that have occurred in the blockchain
     pub chain_graph: chain_graph::ChangeSet<P>,
 }
@@ -147,6 +157,15 @@ impl<K, P> From<chain_graph::ChangeSet<P>> for KeychainChangeSet<K, P> {
     fn from(changeset: chain_graph::ChangeSet<P>) -> Self {
         Self {
             chain_graph: changeset,
+            ..Default::default()
+        }
+    }
+}
+
+impl<K, P> From<DerivationAdditions<K>> for KeychainChangeSet<K, P> {
+    fn from(additions: DerivationAdditions<K>) -> Self {
+        Self {
+            derivation_indices: additions,
             ..Default::default()
         }
     }
--- a/bdk_chain/src/keychain.rs
+++ b/bdk_chain/src/keychain.rs
@@ -21,12 +21,68 @@ mod keychain_txout_index;
 pub use keychain_txout_index::*;
 
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Deserialize, serde::Serialize),
+    serde(
+        crate = "serde_crate",
+        bound(
+            deserialize = "K: Ord + serde::Deserialize<'de>",
+            serialize = "K: Ord + serde::Serialize"
+        )
+    )
+)]
+pub struct DerivationIndices<K>(BTreeMap<K, u32>);
+
+impl<K> DerivationIndices<K> {
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+impl<K: Ord> DerivationIndices<K> {
+    pub fn append(&mut self, mut other: Self) {
+        self.0.iter_mut().for_each(|(key, index)| {
+            if let Some(other_index) = other.0.remove(key) {
+                *index = other_index.max(*index);
+            }
+        });
+
+        self.0.append(&mut other.0);
+    }
+}
+
+impl<K> Default for DerivationIndices<K> {
+    fn default() -> Self {
+        Self(Default::default())
+    }
+}
+
+impl<K> From<BTreeMap<K, u32>> for DerivationIndices<K> {
+    fn from(value: BTreeMap<K, u32>) -> Self {
+        Self(value)
+    }
+}
+
+impl<K> AsRef<BTreeMap<K, u32>> for DerivationIndices<K> {
+    fn as_ref(&self) -> &BTreeMap<K, u32> {
+        &self.0
+    }
+}
+
+impl<K> AsMut<BTreeMap<K, u32>> for DerivationIndices<K> {
+    fn as_mut(&mut self) -> &mut BTreeMap<K, u32> {
+        &mut self.0
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
 /// An update that includes the last active indexes of each keychain.
 pub struct KeychainScan<K, P> {
     /// The update data in the form of a chain that could be applied
     pub update: ChainGraph<P>,
     /// The last active indexes of each keychain
-    pub last_active_indexes: BTreeMap<K, u32>,
+    pub last_active_indexes: DerivationIndices<K>,
 }
 
 impl<K, I> Default for KeychainScan<K, I> {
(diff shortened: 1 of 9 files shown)


evanlinjin:
**Tests to add for `KeychainTxOutIndex`:**

- For a no-wildcard descriptor with no stored scripts:
    - [ ] `next_derivation_index` should return `Some(0)`.
    - [ ] `derive_new` should return `(Some(_), _)`.
    - [ ] `next_unused` should return `(Some(_), _)`.
- For a no-wildcard descriptor with a stored unused script:
    - [ ] `next_derivation_index` should return `None`.
    - [ ] `derive_new` should return `(None, _)`.
    - [ ] `next_unused` should return `(Some(_), _)`.
- For a no-wildcard descriptor with a stored USED script:
    - [ ] `next_unused` should return `(None, _)`.

notmandatory:
I created tests patch here: https://github.com/notmandatory/bdk_core_staging/commit/ea976255ab149b7bf95f091b533acfc5f8f69f6e

If it looks OK feel free to pull or copy/paste in.

If you have suggestions for where else you want new tests I'm happy to do more, this is a great way for me to learn the new core code. Your docs and design make it very easy to understand what's going on! 🙏 This new multi-descriptor feature looks super powerful.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -236,42 +252,37 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-            .inner
-            .script_pubkeys()
-            .get(&index)
-            .expect("we just added it");
-        (next_derivation_index, new_spk)
+        let script = self.inner.spk_at_index(&index).expect("we just stored it");
+        (Some((next_derivation_index, script)), additions)
     }
 
     /// Gets the next usued script pubkey in the keychain i.e. the script pubkey with the lowest index that has not been used yet.
```
rajarshimaitra, line 267: auxiliary typo:
```suggestion
    /// Gets the next unused script pubkey in the keychain i.e. the script pubkey with the lowest index that has not been used yet.
```

rajarshimaitra (commented):
Review ACK 4c563967a60db80e2b1b1150bd59aecb1cc9c62f

below are few nits..

notmandatory:
When I do a `cargo fmt` I get: 
```
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel. 
```
Do we need to turn this off to work with `stable` and our MSRV?

notmandatory:
Added similar tests for wildcard descriptors, if this and prior commit are too verbose let me know and I can tighten it up: https://github.com/notmandatory/bdk_core_staging/commit/d7d18d056da42ead289af97a3aeb281e03853324

rajarshimaitra:
> When I do a cargo fmt I get:

This is happening because `imports_granularity` isn't stable, tracking issue here: https://github.com/rust-lang/rustfmt/issues/4991
So what I do in my local is shift to nightly and then apply `fmt`.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -236,42 +252,37 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-        let new_spk = descriptor
-            .at_derivation_index(next_derivation_index as u32)
-            .derived_descriptor(&secp)
-            .expect("the descriptor cannot need hardened derivation")
-            .script_pubkey();
+    pub fn derive_new(&mut self, keychain: &K) -> (Option<(u32, &Script)>, DerivationAdditions<K>) {
+        let next_derivation_index = match self.next_derivation_index(keychain) {
+            Some(index) => index,
+            None => return (None, DerivationAdditions::default()),
+        };
```
rajarshimaitra, line 259: API question. Is it prudent to return `None` in case of wildcard already stored index? What should the user do to get the address after getting a `None` here? 

I would expect it to give me a script irrespective of derivable or fixed descriptor. for non derivables it would be the same script all the time..

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -156,11 +158,19 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+    /// Get the derivation index after the current one, unless the descriptor has no wildcards.
+    pub fn next_derivation_index(&self, keychain: &K) -> Option<u32> {
+        // we can only get the next index if descriptor exists
+        let has_wildcard = self.keychains.get(keychain)?.has_wildcard();
+
+        match self.derivation_index(keychain) {
+            // for all descriptors, if there is no index, next index is always 0
+            None => Some(0),
+            // if index already exists, and descriptor is not derivable, we cannot get next index
+            Some(_) if !has_wildcard => None,
```
rajarshimaitra, line 170: Sending a `None` back here seems overly verbose, and causes too many cases to keep in mind.  Why not just always return 0 if no wildcard?

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -236,42 +252,37 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
     /// Gets the next usued script pubkey in the keychain i.e. the script pubkey with the lowest index that has not been used yet.
     ///
     /// ## Panics
     ///
     /// Panics if `keychain` has never been added to the index
-    pub fn next_unused(&mut self, keychain: &K) -> (u32, &Script) {
+    pub fn next_unused(
+        &mut self,
+        keychain: &K,
+    ) -> (Option<(u32, &Script)>, DerivationAdditions<K>) {
```
rajarshimaitra, line 275: Same API question as above. What to do when user gets a `None` here?

rajarshimaitra (commented):
few more questions on the new API..

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: I personally think the method should do as advertised. If there are no new derivation indexes to derive, we return `None`.

What does the developer have to "keep in mind" when using this method?

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: I still think the behavior is ideal because it is doing what the method name says it does. If you can give an example how this API can be problematic I would be more convinced.

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: The user knows that there is no more unused. This is excellent information imo.

evanlinjin:
> When I do a `cargo fmt` I get:
> 
> ```
> Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel. 
> ```
> 
> Do we need to turn this off to work with `stable` and our MSRV?

It wasn't me who decided to use unstable features of `cargo fmt` and I don't have an opinion on this. @LLFourn What do you think?

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: I just had a talk with @nickfarrow to get a second opinion. We concluded that it is safer to return `None`.

Imagine a situation when there is an accidental inclusion of a non-wilcard descriptor, and the API for `derive_new` already returns something. This can risk address-reuse.

Also, even descriptors with wildcards are limited in how many scripts it can derive (also I haven't tested this).

Thinking from the other direction, it would be trivial to implement the logic you suggested on top of an API that can return `None`.

bdk_chain/src/keychain.rs
```diff
@@ -194,20 +262,32 @@ mod test {
...
+            lhs.derivation_indices.as_ref().get(&Keychain::Two),
+            Some(&5)
+        );
+        assert_eq!(
+            lhs.derivation_indices.as_ref().get(&Keychain::Three),
+            Some(&3)
+        );
+        assert_eq!(
+            lhs.derivation_indices.as_ref().get(&Keychain::Four),
+            Some(&4)
```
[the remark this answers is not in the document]
evanlinjin: Created an issue.

bdk_chain/src/keychain/keychain_txout_index.rs
LLFourn: I think that it is useful information but can't we communicate this with an empty changeset?

bdk_chain/src/keychain/keychain_txout_index.rs
LLFourn: I'm leaning towards @rajarshimaitra view. `None` is more dangerous than returning a reused address since people will be inclined to unwrap it and open up a panic vector that could be exploitable or become exploitable later on. The information that you were given a stale address can be seen by calling the `DerivationAddtions::is_empty`.

bdk_chain/src/keychain/keychain_txout_index.rs
LLFourn: I think it'd be nice to return a `(u32, bool)` to communicate this in the case you agree with me that it is sufficient to communicate it with the `DerivationAdditions` in the other methods

bdk_chain/src/keychain.rs
```diff
@@ -20,13 +20,79 @@ mod keychain_txout_index;
...
+    serde(
+        crate = "serde_crate",
+        bound(
+            deserialize = "K: Ord + serde::Deserialize<'de>",
+            serialize = "K: Ord + serde::Serialize"
+        )
+    )
+)]
+#[must_use]
+pub struct DerivationAdditions<K>(BTreeMap<K, u32>);
```
LLFourn, line 36: Won't this not necessarily be monotone when you add the trim methods in #104. Maybe just call it a changeset too.

Actually what's wrong with just a `BTreeMap` here? I don't see the newtype doing much here.

bdk_chain/tests/test_keychain_tracker.rs
```diff
@@ -74,7 +74,13 @@ fn test_balance() {
...
         output: vec![TxOut {
             value: 13_000,
-            script_pubkey: tracker.txout_index.derive_new(&Keychain::One).1.clone(),
+            script_pubkey: tracker
+                .txout_index
+                .derive_new(&Keychain::One)
+                .0
+                .unwrap()
+                .1
+                .clone(),
```
LLFourn, line 83: Why not just put `DerivationAdditions` inside the `Option`? Much less friction that way and you don't care about it in the case where the option is `None`.

bdk_chain/src/keychain.rs
```diff
@@ -20,13 +20,79 @@ mod keychain_txout_index;
...
+    derive(serde::Deserialize, serde::Serialize),
+    serde(
+        crate = "serde_crate",
+        bound(
+            deserialize = "K: Ord + serde::Deserialize<'de>",
+            serialize = "K: Ord + serde::Serialize"
+        )
+    )
+)]
+#[must_use]
```
LLFourn, line 35: Does this work when the thing is inside a tuple?

bdk_chain/src/keychain/keychain_txout_index.rs
rajarshimaitra: > What does the developer have to "keep in mind" when using this method?

That "if my non-derivable descriptors are already derived once, then there's a `None` case to handle, while for derivables it's `Some(_)`". 

IMO this forces the user to write wallet logic differently for deliverables and non-wildcard. Which wasn't the case previously in the API. So my question was why not just keep the previous API intact? 

> Imagine a situation when there is an accidental inclusion of a non-wilcard descriptor, and the API for derive_new already returns something. This can risk address-reuse.

Sometimes there are non-wildcard descriptors that are intended to have address reuses. The current BDK API is also not opinionated on descriptor types. If a user puts in a non-wildcard it is assumed that he knows it's gonna be the same address.

> Thinking from the other direction, it would be trivial to implement the logic you suggested on top of an API that can return None.

Agreed.. It's totally trivial to abstract away in a higher layer. 

I mean, if I wanna handle non-derivable differently from derivable I can just call `has_wildcard()` and craft my logic accordingly. Is the information here worth the complexity of the resultant API?

bdk_chain/src/keychain.rs
evanlinjin: The `#[must_use]`? Or the derives?

bdk_chain/src/keychain.rs
evanlinjin: > Actually what's wrong with just a BTreeMap here? I don't see the newtype doing much here.

I wanted to have the `append` logic just like the rest of our changeset/additions. I feel like it would be an extremely useful API to have.

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: > I think it'd be nice to return a `(u32, bool)` to communicate this in the case you agree with me that it is sufficient to communicate it with the `DerivationAdditions` in the other methods

@LLFourn However, we aren't mutating things here. I think `(u32, bool)` makes sense.

bdk_chain/src/keychain.rs
LLFourn: `BTreeMap` has `.append` though.

evanlinjin: Not the type of "append" we need though. We need to keep the highest value, not only set the ones that are missing.

rajarshimaitra:
@evanlinjin made the modification in an alternate PR https://github.com/LLFourn/bdk_core_staging/pull/154

bdk_chain/src/keychain.rs
evanlinjin: I assume so...

PR #152, opened 2023-01-17
Remove rustfmt.toml

LLFourn:
and it probably doesn't make sense anymore.


evanlinjin (approved):
ACK 3b7aed7dae41ec5caa5543903c449d4dbb3a9500

PR #153, opened 2023-01-19
Add example with compact block filters

danielabrozzoni:


evanlinjin:
I just had a brief look through, nothing seems apparent to me why balance would disappear after restart. Will properly debug later.

danielabrozzoni:
Rebased on top of #157 

(Still doesn't work though)

evanlinjin:
> Rebased on top of #157 
> 
> (Still doesn't work though)

Did you also add on the fix mentioned in Discord?

evanlinjin:
https://github.com/danielabrozzoni/bdk_core_staging/pull/2

I did a PR into yours fixing the balance disappearing after restart.

Looking good so far. In my opinion, the next steps for this PR:

Don't hold a mutable reference of the tracker when syncing. We need a structure (imo, like an iterator), where we only grab initial data from the tracker to create. Then it spits out data which the tracker can consume.

In my head, this thing that the CBF Iterator spits out can look something like this:

```rust
pub enum CbfData {
    RelevantBlock(bitcoin::Block),
    DisconnectedBlock(u32), // block height
}
```

danielabrozzoni:
> Don't hold a mutable reference of the tracker when syncing. We need a structure (imo, like an iterator), where we only grab initial data from the tracker to create. Then it spits out data which the tracker can consume.

Right, done! :)

> Additionally, we need to somehow figure out how to properly do stop gap. We need logic that goes "uhh oh, we need more scripts for a rescan"!

I think this is pretty trivial to implement, but I'd prefer to do it after receiving some feedback on the current code, if possible

bdk_cbf_example/src/cbf.rs
```diff
@@ -0,0 +1,166 @@
...
+        Ok(CbfClient {
+            handle,
+            setup: None,
+        })
+    }
+
+    pub fn sync_setup(
+        &mut self,
+        scripts: impl Iterator<Item = Script>,
+        processed_height: Option<BlockId>,
```
evanlinjin, line 73: Instead of `processed_height` as an input, what do you think about using `local_chain: BTreeMap<u32, BlockHash>` and iterate in reverse until we find the last point of agreement?

LLFourn: Nakamoto has `handle::Handle::get_block_by_height` so this should be doable.

bdk_cbf_example/src/cbf.rs
```diff
@@ -0,0 +1,166 @@
...
+                            println!("* peer negotiated with height {:?}", height);
+                            let height = height as u32;
+                            if setup.peer_height < Some(height) {
+                                setup.peer_height = Some(height);
+                            }
+                            if setup.processed_height.map(|h| h.height) == setup.peer_height && setup.blocks_matched.is_empty() {
+                                // TODO: improve this!
+                                // It might be that both me and this peer
+                                // are lagging behind
+                                return Ok(CbfEvent::ChainSynced(setup.processed_height));
```
LLFourn, line 118: What about using nakamoto's `Event::Synced` to emit this? This seems very error prone and even an attack vector. I think you shouldn't need to keep track of `setup.peer_height` at all.

So when you get `Event::Synced` with height == `Handle::get_tip()` you can emit this.

bdk_cbf_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+                .into_values()
+                .flatten()
+                .map(|(_, spk)| spk.clone());
+
+            let last_checkpoint = keychain_tracker.chain_graph().chain().latest_checkpoint();
+
+            let mut update = ChainGraph::<TxHeight>::default();
+
+            if let Some(last_cp) = last_checkpoint.clone() {
+                // Otherwise the chains don't connect :)
```
LLFourn, line 105: what if they don't connect though? :)

bdk_cbf_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+                match client.next_event()? {
+                    CbfEvent::BlockMatched(id, txs) => {
+                        let _ = update.insert_checkpoint(BlockId {
+                            height: id.height as u32,
+                            hash: id.hash,
+                        })?;
+
+                        for (tx, height) in txs {
+                            //keychain_tracker.lock().unwrap().txout_index.pad_all_with_unused(stop_gap);
+                            if keychain_tracker.txout_index.is_relevant(&tx) {
```
LLFourn, line 122: I think rather than doing this here we should simply have a way of filtering out irrelevant things out of an update. For now we can just include all transactions in the block. I made an issue to build the tools to do this properly: https://github.com/LLFourn/bdk_core_staging/issues/193

There really isn't much harm to just acquiring a `.lock()` on the db and the keychain tracker and just doing an `apply_update` to the `KeychainTracker` and pushing the thing into the database right away. This is good because it saves your progress for each block as you go.

bdk_cbf_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+                                println!("* adding tx to update: {} @ {}", tx.txid(), height);
+                                let _ = update.insert_tx(tx.clone(), height)?;
+                            }
+                            derivation_additions.append(keychain_tracker.txout_index.scan(&tx));
+                        }
+                    }
+                    CbfEvent::BlockDisconnected(id) => {
+                        // TODO: what happens if a block gets disconnected before I process its
+                        // filter?
```
LLFourn, line 131: I really hope that `BlockMatched` doesn't get emitted for things that are not in the chain! Hopefully nakamoto guarantees this.

bdk_cbf_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+        CbfCommands::Rescan { _batch_size } => {
+            todo!("Implement rescan from sync");
+            // This function will reveal batch_size scripts and rescan the nakamoto
+            // client
+            // If we notice, after rescan, that more than `batch_size - epsilon` scripts
+            // have been used, we rescan using 2*batch_size, and so on
+        }
+        CbfCommands::Scan => {
+            // indexing logic!
+            let mut keychain_tracker = keychain_tracker.lock().unwrap();
```
LLFourn, line 90: Try and do every `lock` in scoped block `{...}` so the reader can see where locks can be taken and for how long they need to be taken for. Here you should only need the lock for the time it takes to get the scripts you're interested in and the checkpoints. You can return them from the block and release the lock.

bdk_cbf_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+                    }
+                }
+            }
+
+            let changeset = KeychainChangeSet {
+                derivation_indices: derivation_additions,
+                chain_graph: keychain_tracker
+                    .chain_graph()
+                    .determine_changeset(&update)?,
+            };
```
LLFourn, line 152: We should aim to not do this manually after #193 is done.

bdk_cli_lib/src/lib.rs
```diff
@@ -19,14 +19,15 @@ use bdk_chain::{
 use bdk_coin_select::{coin_select_bnb, CoinSelector, CoinSelectorOpt, WeightedValue};
 pub use clap;
 use clap::{Parser, Subcommand};
+pub use log;
```
LLFourn, line 22: Not sure if we need this. I think CLI should try and avoid using logging unless we're going into a kind of `daemon` mode. I think env_logger is a good idea though but I don't see it being initialized anywhere yet.

bdk_cbf_example/src/main.rs
```diff
@@ -0,0 +1,158 @@
...
+use bdk_chain::keychain::{DerivationAdditions, KeychainChangeSet};
+use bdk_chain::{BlockId, TxHeight};
+
+mod cbf;
+
+#[derive(Args, Debug, Clone)]
+struct CbfArgs {
+    #[arg(value_enum, default_value_t = Domains::IPv4)]
+    domains: Domains,
+}
```
LLFourn, line 17: I'd prefer it if `domains` were just arguments to the actual commands that use `domains` and to not have `CbfArgs`. FWIW I've complained about adding global arguments to the `bdk_cli::init` to @evanlinjin but it looks like he didn't remove it from that PR yet.

LLFourn (commented):
Thanks @danielabrozzoni. I'm excited about nakamoto CBF. It's looking like a great option. I think we should do https://github.com/LLFourn/bdk_core_staging/issues/193 so we can simplify the implementation here.

evanlinjin:
@LLFourn I've tried removing the global CLI args but I couldn't find a way to do it. The "general commands" need to take in a client, and so the client needs to be initiated before our chain-specific commands.

LLFourn:
@evanlinjin I see so you actually need to know "domains" for broadcasting too. So let's add the ability to add params to `Send` (rather than globally). Then pass those params through to the closure I made in #198 for broadcasting. Then you can create the client inside the closure.

PR #154, opened 2023-01-21
[Alternate] Methods that change derivation indicies should return changesets

### LINKED ISSUE #151: Add more cases for `append_keychain_derivation_indices` test
We can cover one more case of `rhs` having keychain `Five` and `Six`, and after appending `lhs` should have the new keychains too.

_Originally posted by @rajarshimaitra in https://github.com/LLFourn/bdk_core_staging/pull/150#discussion_r1071400266_

rajarshimaitra:
This is a parallel of #150 . I have approached the changes as discussed in the PR thread. 

Fixes #151 

Added more Docs.

Todo: More tests. I am already working on them. Will push soon. 

Opened for early comments.

### DIFF
--- a/bdk_chain/src/keychain.rs
+++ b/bdk_chain/src/keychain.rs
@@ -20,6 +20,11 @@ mod keychain_txout_index;
 #[cfg(feature = "miniscript")]
 pub use keychain_txout_index::*;
 
+/// Represents updates to the `last_active` for each `keychain`.
+///
+/// TODO: In the future, we will add an additional index per keychain for "lookahead". This will be
+/// helpful for syncing/scanning data from blockchains without increasing the user-derived index
+/// count.
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(
     feature = "serde",
@@ -33,11 +38,15 @@ pub use keychain_txout_index::*;
     )
 )]
 #[must_use]
-pub struct DerivationAdditions<K>(BTreeMap<K, u32>);
+pub struct DerivationAdditions<K> {
+    /// The updates (if any) to the last-derived index per keychain.
+    pub last_derived: BTreeMap<K, u32>,
+}
 
 impl<K> DerivationAdditions<K> {
+    /// Returns `true` if the additions are empty.
     pub fn is_empty(&self) -> bool {
-        self.0.is_empty()
+        self.last_derived.is_empty()
     }
 }
 
@@ -47,19 +56,21 @@ impl<K: Ord> DerivationAdditions<K> {
     /// If keychain already exists, increases the index, if other's index > self's index
     /// If keychain didn't exist, appends the new keychain
     pub fn append(&mut self, mut other: Self) {
-        self.0.iter_mut().for_each(|(key, index)| {
-            if let Some(other_index) = other.0.remove(key) {
+        self.last_derived.iter_mut().for_each(|(key, index)| {
+            if let Some(other_index) = other.last_derived.remove(key) {
                 *index = other_index.max(*index);
             }
         });
 
-        self.0.append(&mut other.0);
+        self.last_derived.append(&mut other.last_derived);
     }
 }
 
 impl<K> Default for DerivationAdditions<K> {
     fn default() -> Self {
-        Self(Default::default())
+        Self {
+            last_derived: Default::default(),
+        }
     }
 }
 
@@ -68,19 +79,9 @@ where
     I: IntoIterator<Item = (K, u32)>,
 {
     fn from(value: I) -> Self {
-        Self(value.into_iter().collect())
-    }
-}
-
-impl<K> AsRef<BTreeMap<K, u32>> for DerivationAdditions<K> {
-    fn as_ref(&self) -> &BTreeMap<K, u32> {
-        &self.0
-    }
-}
-
-impl<K> AsMut<BTreeMap<K, u32>> for DerivationAdditions<K> {
-    fn as_mut(&mut self) -> &mut BTreeMap<K, u32> {
-        &mut self.0
+        Self {
+            last_derived: value.into_iter().collect(),
+        }
     }
 }
 
@@ -282,22 +283,22 @@ mod test {
 
         // Exiting index doesn't update if new index in `other` is lower than `self`
         assert_eq!(
-            lhs.derivation_indices.as_ref().get(&Keychain::One),
+            lhs.derivation_indices.last_derived.get(&Keychain::One),
             Some(&7)
         );
         // Existing index updates if new index in `other` is higher than `self.
         assert_eq!(
-            lhs.derivation_indices.as_ref().get(&Keychain::Two),
+            lhs.derivation_indices.last_derived.get(&Keychain::Two),
             Some(&5)
         );
         // Existing index unchanged, if keychain doesn't exist in `other`
         assert_eq!(
-            lhs.derivation_indices.as_ref().get(&Keychain::Three),
+            lhs.derivation_indices.last_derived.get(&Keychain::Three),
             Some(&3)
         );
         // New keychain gets added if keychain is in `other`, but not in `self`.
         assert_eq!(
-            lhs.derivation_indices.as_ref().get(&Keychain::Four),
+            lhs.derivation_indices.last_derived.get(&Keychain::Four),
             Some(&4)
         );
     }
--- a/bdk_chain/src/keychain.rs
+++ b/bdk_chain/src/keychain.rs
@@ -63,12 +63,6 @@ impl<K> Default for DerivationAdditions<K> {
     }
 }
 
-// impl<K> From<BTreeMap<K, u32>> for DerivationAdditions<K> {
-//     fn from(value: BTreeMap<K, u32>) -> Self {
-//         Self(value)
-//     }
-// }
-
 impl<K: Ord, I> From<I> for DerivationAdditions<K>
 where
     I: IntoIterator<Item = (K, u32)>,
--- a/bdk_chain/src/keychain.rs
+++ b/bdk_chain/src/keychain.rs
@@ -21,12 +21,82 @@ mod keychain_txout_index;
 pub use keychain_txout_index::*;
 
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Deserialize, serde::Serialize),
+    serde(
+        crate = "serde_crate",
+        bound(
+            deserialize = "K: Ord + serde::Deserialize<'de>",
+            serialize = "K: Ord + serde::Serialize"
+        )
+    )
+)]
+#[must_use]
+pub struct DerivationAdditions<K>(BTreeMap<K, u32>);
+
+impl<K> DerivationAdditions<K> {
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+impl<K: Ord> DerivationAdditions<K> {
+    /// Append another [`DerivationAdditions`] into self.
+    ///
+    /// If keychain already exists, increases the index, if other's index > self's index
+    /// If keychain didn't exist, appends the new keychain
+    pub fn append(&mut self, mut other: Self) {
+        self.0.iter_mut().for_each(|(key, index)| {
+            if let Some(other_index) = other.0.remove(key) {
+                *index = other_index.max(*index);
+            }
+        });
+
+        self.0.append(&mut other.0);
+    }
+}
+
+impl<K> Default for DerivationAdditions<K> {
+    fn default() -> Self {
+        Self(Default::default())
+    }
+}
+
+// impl<K> From<BTreeMap<K, u32>> for DerivationAdditions<K> {
+//     fn from(value: BTreeMap<K, u32>) -> Self {
+//         Self(value)
+//     }
+// }
+
+impl<K: Ord, I> From<I> for DerivationAdditions<K>
+where
+    I: IntoIterator<Item = (K, u32)>,
+{
+    fn from(value: I) -> Self {
+        Self(value.into_iter().collect())
+    }
+}
+
+impl<K> AsRef<BTreeMap<K, u32>> for DerivationAdditions<K> {
+    fn as_ref(&self) -> &BTreeMap<K, u32> {
+        &self.0
+    }
+}
+
+impl<K> AsMut<BTreeMap<K, u32>> for DerivationAdditions<K> {
+    fn as_mut(&mut self) -> &mut BTreeMap<K, u32> {
+        &mut self.0
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
 /// An update that includes the last active indexes of each keychain.
 pub struct KeychainScan<K, P> {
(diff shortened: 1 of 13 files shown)


rajarshimaitra:
Added test coverage for the new APIs. Commented out the test cases at numeric bound as naturally, that will gulp up all the rams.. Still included for documentation purposes.

@evanlinjin @LLFourn, this I think is now ready for a look..

evanlinjin (commented):
Thanks for the PR! I only had a brief look, but it looks good so far.

@LLFourn do you think panicking when a keychain does not exist is the right behavior to have?

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -156,11 +158,40 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+        match self.derivation_index(keychain) {
+            // for all descriptors, if there is no index, next_index is always 0
+            None => (0, true),
+            // for derivable descriptors, we get the next index (without overflow)
+            // for un-derivable descriptors we always get next_index = 0
+            Some(index) => {
+                if !has_wildcard {
+                    (0, false)
+                } else {
+                    if let Some(new_index) = index.checked_add(1) {
```
evanlinjin, line 187: Do we also need to check bounds at `2^31`? It might be more convenient to create a constant.

evanlinjin: Done in d488f4eb250de16074ef8c0b44acb1a64a28ef06

evanlinjin:
@rajarshimaitra I did a fixup to the documentation, and implemented a fix for `store_up_to` (as it panicked when `up_to` was greater than derivation bounds).

evanlinjin (approved):
ACK d488f4eb250de16074ef8c0b44acb1a64a28ef06

bdk_chain/tests/test_keychain_txout_index.rs
```diff
@@ -56,14 +60,9 @@ fn test_store_all_up_to() {
...
 fn test_pad_all_with_unused() {
-    let mut txout_index = init_txout_index();
+    let (mut txout_index, external_desc, _) = init_txout_index();
 
-    let external_spk3 = txout_index
-        .keychains()
-        .get(&TestKeychain::External)
-        .unwrap()
-        .at_derivation_index(3)
-        .script_pubkey();
```
rajarshimaitra, line 45: IMO this code was useful.. It shows how to get the keychains from a tracker..

rajarshimaitra (commented):
Review ACK d488f4eb250de16074ef8c0b44acb1a64a28ef06

Overall it looks good..

Below are few nits on docs.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -75,7 +77,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     ///
     /// 1. After loading transaction data from disk you may scan over all the txouts to restore all
     /// your txouts.
-    /// 2. When getting new data from the chain you usually scan it before incoporating it into your chain state.
+    /// 2. When getting new data from the chain you usually scan it before incoporating it into your
+    /// chain state.
```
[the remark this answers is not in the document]
evanlinjin: > What does `chain state` implies here? We are adding data into the index with this method. So I was confused as it first seemed like we are talking about the sparsechain. Maybe call it something more clear like the `tracker` or the `indexes`? 

We are taking about the sparsechain here. The sparsechain should only contain relevant transactions.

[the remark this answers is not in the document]
rajarshimaitra: Then better to call it `sparsechain` directly?

bdk_chain/tests/test_keychain_txout_index.rs
evanlinjin: But this is a test, not an example

LLFourn (commented):
This looks fantastic. I did a lot of critical thinking on this trying to find a way this approach could go wrong and in the end came to the conclusion that it's right approach. I'm going to push a commit that makes a few minor changes:

1. We're missing an `apply_additions` method on the index
2. I think the example is not quite using the new API correctly (or the old one). I think we should ensure we don't keep a changeset "open" over an IO request as we need to guarantee that changesets are not applied out of order (although here I'm not sure if it can cause much damage).

I've left a few nit comments.

LLFourn:
I made #159 and now don't have time to do the `apply_additions` method. If anyone wants to finish that for me go ahead otherwise I'll try and get to it later.

evanlinjin:
> I made #159 and now don't have time to do the `apply_additions` method. If anyone wants to finish that for me go ahead otherwise I'll try and get to it later.

Attempted in #160

bdk_chain/src/keychain.rs
```diff
@@ -20,13 +20,77 @@ mod keychain_txout_index;
...
+}
+
+impl<K: Ord, I> From<I> for DerivationAdditions<K>
+where
+    I: IntoIterator<Item = (K, u32)>,
+{
+    fn from(value: I) -> Self {
+        Self(value.into_iter().collect())
+    }
+}
```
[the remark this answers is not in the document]
evanlinjin: Unfortunately this is needed for the tests where we do want to manually construct changesets/additions. I don't have a better solution.

bdk_chain/src/keychain.rs
```diff
@@ -20,13 +20,77 @@ mod keychain_txout_index;
...
+    fn as_ref(&self) -> &BTreeMap<K, u32> {
+        &self.0
+    }
+}
+
+impl<K> AsMut<BTreeMap<K, u32>> for DerivationAdditions<K> {
+    fn as_mut(&mut self) -> &mut BTreeMap<K, u32> {
+        &mut self.0
+    }
+}
```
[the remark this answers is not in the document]
evanlinjin: We could make the internal a `pub(crate)`, but we still need to change the esplora example.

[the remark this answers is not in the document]
LLFourn: Ahhh `last_active_indexes` in `KeychainScan` should not be `DerivationAdditions`. That's a mistake.

bdk_chain/src/keychain.rs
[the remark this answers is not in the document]
LLFourn: We don't want to manually construct them in the tests. We just want to check they have the right values inside but you can do that with `as_ref` (we should probably have `as_inner` as a member function as well).

bdk_chain/src/keychain.rs
```diff
@@ -20,6 +20,67 @@ mod keychain_txout_index;
...
+        Self(Default::default())
+    }
+}
+
+impl<K> AsRef<BTreeMap<K, u32>> for DerivationAdditions<K> {
+    fn as_ref(&self) -> &BTreeMap<K, u32> {
+        &self.0
+    }
+}
+
```
evanlinjin, line 83: Since we have `as_inner` now, I think we can remove this.

bdk_chain/src/keychain/keychain_tracker.rs
```diff
@@ -83,7 +83,7 @@ where
 
     pub fn apply_changeset(&mut self, changeset: KeychainChangeSet<K, P>) {
         self.txout_index
-            .store_all_up_to(&changeset.derivation_indices);
+            .apply_additions(changeset.derivation_indices.clone());
         self.txout_index.scan(&changeset);
         self.chain_graph.apply_changeset(changeset.chain_graph)
     }
```
evanlinjin, line 89:
```rust
    pub fn apply_changeset(&mut self, changeset: KeychainChangeSet<K, P>) {
        let KeychainChangeSet {
            derivation_indices,
            chain_graph,
        } = changeset;
        self.txout_index.apply_additions(derivation_indices);
        self.txout_index.scan(&chain_graph);
        self.chain_graph.apply_changeset(chain_graph)
    }
```

This way there is no unnecessary clones.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -172,140 +206,167 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
 
-    /// Gets the current derivation index for each keychain in the index.
+    /// Get the current derivation index for each keychain.
+    ///
+    /// Keychains with no indicies derived will not be included in the returned [`BTreeMap`].
     pub fn derivation_indices(&self) -> BTreeMap<K, u32> {
         self.keychains()
             .keys()
             .filter_map(|keychain| Some((keychain.clone(), self.derivation_index(&keychain)?)))
             .collect()
```
evanlinjin, line 216: I think it would make sense to return an iterator here, instead of collecting.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -172,140 +206,146 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-    ///
-    /// [`pad_with_unused`]: Self::pad_with_unused
-    pub fn pad_all_with_unused(&mut self, pad_len: u32) -> bool {
-        let mut changed = false;
-        let keychains = self
-            .keychains
-            .keys()
-            .cloned()
-            .collect::<alloc::vec::Vec<_>>();
+    pub fn pad_all_with_unused(&mut self, pad_len: u32) -> DerivationAdditions<K> {
```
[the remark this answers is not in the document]
evanlinjin: @LLFourn Is there a ticket yet?

bdk_chain/src/spk_txout_index.rs
```diff
@@ -197,6 +197,30 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
...
+    ///
+    /// Note that if `self` has scanned an output with this script pubkey then this will have no
+    /// effect.
+    ///
+    /// [`mark_used`]: Self::mark_used
+    pub fn unmark_used(&mut self, index: &I) {
+        if self.outputs_in_range(index..=index).next().is_none() {
+            self.unused.insert(index.clone());
+        }
+    }
```
evanlinjin, line 222: I think it will be helpful to return a `bool` for these methods which is `true` when changes are made.

bdk_cli_lib/src/lib.rs
```diff
@@ -336,7 +352,7 @@ pub fn create_tx<P: ChainPosition>(
 
     let mut change_output = TxOut {
         value: 0,
-        script_pubkey: change_script,
+        script_pubkey: change_script.clone(),
```
evanlinjin, line 355:
```suggestion
        script_pubkey: change_script,
```

bdk_chain/src/spk_txout_index.rs
evanlinjin: We also need to make sure the `index` exists in `script_pubkeys` before `unmark_used`. I also think `mark_unused` is a better name than `unmark_used`.

evanlinjin: Fixed in e596ffcef7e701a05c0e426dec6217d1ca217aa1

bdk_cli_lib/src/lib.rs
evanlinjin: Fixed in ee31e22d9b80e0ad33c43d66e3271b3ebb90a7e3

bdk_chain/src/keychain/keychain_tracker.rs
evanlinjin: Fixed in ee31e22d9b80e0ad33c43d66e3271b3ebb90a7e3

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Defer.

bdk_cli_lib/src/lib.rs
```diff
@@ -478,42 +502,89 @@ where
...
+                } else {
+                    (transaction, None)
+                }
+            };
+
+            match client.broadcast(&transaction) {
+                Ok(_) => {
+                    println!("Broadcasted Tx : {}", transaction.txid());
+                    let mut tracker = tracker.lock().unwrap();
+                    match tracker.insert_tx(transaction.clone(), P::unconfirmed()) {
```
evanlinjin, line 546: I think it would make more sense to try insert it into `tracker` prior to broadcasting and directly after the transaction is created. This way, we can check the created transaction against our `ChainGraph` rules.

If broadcast fails because the transaction cannot be accepted into the mempool, we can have a method to remove unconfirmed transaction from `SparseChain`/`ChainGraph`, and signal to the user that we need a sync or scan.

If broadcast fails because of network issues, we can retry until a timeout.

LLFourn: > I think it would make more sense to try insert it into `tracker` prior to broadcasting and directly after the transaction is created. This way, we can check the created transaction against our `ChainGraph` rules.

Isn't this what `insert_tx_preview` is for?

We can't insert it before unless we are either (i) willing to save it to db at that time OR (ii) keep a lock on the db while we broadcast.  (i) is bad because then the change address will always be marked as used (txout index will mark it as such when it sees it come on from DB) and (ii) is bad because it keeps a potentially long lock on the db.

bdk_cli_lib/src/lib.rs
```diff
@@ -163,27 +166,30 @@ pub struct AddrsOutput {
...
     KeychainChangeSet<Keychain, P>: serde::Serialize + serde::de::DeserializeOwned,
 {
-    let txout_index = &mut keychain_tracker.txout_index;
+    let mut tracker = tracker.lock().unwrap();
+    let txout_index = &mut tracker.txout_index;
 
-    let new_address = match addr_cmd {
+    let addr_cmmd_output = match addr_cmd {
         AddressCmd::Next => Some(txout_index.next_unused(&Keychain::External)),
         AddressCmd::New => Some(txout_index.derive_new(&Keychain::External)),
```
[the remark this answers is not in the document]
LLFourn: Why do you want to mark unused addresses as used?

bdk_chain/src/keychain/keychain_txout_index.rs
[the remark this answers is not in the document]
LLFourn: #163

bdk_cli_lib/src/lib.rs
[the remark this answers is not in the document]
evanlinjin: When choosing a change spk, we use the "next unused", but if the change and external keychains is the same keychain, the "next unused" might have already been given out.

bdk_cli_lib/src/lib.rs
evanlinjin: > Isn't this what insert_tx_preview is for?

evanlinjin: I still think (i) is ideal. We would want to keep the change address marked as used since we have already potentially exposed it (via broadcast).

LLFourn: but what if we know we have not exposed it? i.e. broadcast failed in a way that does not expose it.

PR #155, opened 2023-01-23
RPC chain example

danielabrozzoni:
This PR rebases and fixes compilation of #89

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ Cargo.lock
 out.perf-folded
 perf.data
 rust-perf.svg
+
 .bdk_example_db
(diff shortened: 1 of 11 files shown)


evanlinjin:
Thanks @danielabrozzoni ! Just skimmed through it now, I realized there were some changes to the other modules in this PR that we don't actually want to include anymore :grimacing: 

This is a boring task, maybe I can take over?

danielabrozzoni:
>  I realized there were some changes to the other modules in this PR that we don't actually want to include anymore grimacing

Mhh, I'm using the same structures in the cbf examples, so I'll need to follow this one anyways to see how to update the cbf :)

> This is a boring task, maybe I can take over?

As you wish!

evanlinjin:
Replaced by #157

PR #157, opened 2023-01-27
RPC chain example

evanlinjin:
Replaces #155 and #89


danielabrozzoni:
I pushed a rebase of this PR in https://github.com/danielabrozzoni/bdk_core_staging/tree/rpc

Double-check if it's alright: it's using the new methods of the KeychainTxOutIndex, and while I'm fairly confident I'm using them properly, don't trust and verify :)

PR #160, opened 2023-02-01
Implement `KeychainTxOutIndex::apply_additions`

evanlinjin:
Internally, this just uses `store_all_up_to`, however this will change once we start storing a "lookahead" index for each keychain.

Additionally, I have removed `AsRef` and `AsMut` implementations for ` DerivationAdditions`, and `DerivationAdditions` now has a single field `last_derived`.

---

This is my attempt at addressing https://github.com/LLFourn/bdk_core_staging/pull/154#issuecomment-1411479890

### DIFF
--- a/bdk_chain/src/keychain.rs
+++ b/bdk_chain/src/keychain.rs
@@ -20,6 +20,11 @@ mod keychain_txout_index;
 #[cfg(feature = "miniscript")]
 pub use keychain_txout_index::*;
 
+/// Represents updates to the `last_active` for each `keychain`.
+///
+/// TODO: In the future, we will add an additional index per keychain for "lookahead". This will be
+/// helpful for syncing/scanning data from blockchains without increasing the user-derived index
+/// count.
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(
     feature = "serde",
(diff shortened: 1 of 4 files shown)


bdk_chain/src/keychain.rs
```diff
@@ -33,11 +38,15 @@ pub use keychain_txout_index::*;
     )
 )]
 #[must_use]
-pub struct DerivationAdditions<K>(BTreeMap<K, u32>);
+pub struct DerivationAdditions<K> {
+    /// The updates (if any) to the last-derived index per keychain.
+    pub last_derived: BTreeMap<K, u32>,
```
LLFourn, line 43: I didn't see what was wrong with the way it was before. It could use a method `as_inner` or something that returns an immutable ref to the `BTreeMap`.

bdk_chain/src/keychain.rs
```diff
@@ -20,6 +20,11 @@ mod keychain_txout_index;
 #[cfg(feature = "miniscript")]
 pub use keychain_txout_index::*;
 
+/// Represents updates to the `last_active` for each `keychain`.
+///
+/// TODO: In the future, we will add an additional index per keychain for "lookahead". This will be
+/// helpful for syncing/scanning data from blockchains without increasing the user-derived index
+/// count.
```
LLFourn, line 27: Remove this TODO because this won't apply to DerivationAdditions.

LLFourn (changes requested):

LLFourn:
> Internally, this just uses `store_all_up_to`, however this will change once we start storing a "lookahead" index for each keychain.

I imagine that we will delete `store_all_up_to` and `apply_additions` will just take all its logic.

LLFourn:
Can you make this PR to #154 instead?

LLFourn:
I merged this into #154 already

PR #169, opened 2023-02-06
Introduce `lookahead` for `KeychainTxOutIndex`

### LINKED ISSUE #163: KeychainTxOutIndex should look ahead
Doing "stop gap" like logic when scanning through blocks is annoying with the current API because we conflate with the "index which I will give out next address I will give out" with the "index I derive scripts up to so I can find outputs for them". Just like `bdk` already does we can have a kind of look ahead to make this easier.

The API should be:

- `set_lookahead(keychain: &K, look_head: u32)
- `set_all_lookaheads(look_ahead: u32)`
- `lookaheads(&self) -> &BTreeMap<K, u32>`

Then the `KeychainTxOutIndex` should just make sure that it is has always inserted `lookahead` more spks into the inner `SpkTxOutIndex`. `scan` will also need to return the updated `DerivationAdditions` for each call since now scannig can increase your derivation index.

We can also remove `store_*` methods from the API and rebrand them to `set_derivation_index` which should also return a `DerivationAdditions`.

evanlinjin:
closes #163 

The `lookahead` is the number of scripts to cache ahead of the last derived script index.

Scanning methods now return `DerivationAdditions` and replenish the internal `lookahead`.

### TODO

- [x] Add `lookaheads(&self) -> &BTreeMap<K, u32>`
- [x] Rebrand `store..` methods as `set_derivation_index`
- [x] Better documentation
- [x] Tests

### DIFF
--- a/bdk_chain/src/keychain/keychain_txout_index.rs
+++ b/bdk_chain/src/keychain/keychain_txout_index.rs
@@ -107,7 +107,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// [`ForEachTxout`]: crate::ForEachTxout
     pub fn scan(&mut self, txouts: &impl ForEachTxout) -> DerivationAdditions<K> {
         let mut additions = DerivationAdditions::<K>::default();
-        txouts.for_each_txout(&mut |(op, txout)| additions.append(self.scan_txout(op, txout)));
+        txouts.for_each_txout(|(op, txout)| additions.append(self.scan_txout(op, txout)));
         additions
     }
 
--- a/bdk_chain/src/keychain/keychain_txout_index.rs
+++ b/bdk_chain/src/keychain/keychain_txout_index.rs
@@ -3,7 +3,11 @@ use crate::{
     miniscript::{Descriptor, DescriptorPublicKey},
     ForEachTxout, SpkTxOutIndex,
 };
-use bitcoin::{secp256k1::Secp256k1, OutPoint, Script, TxOut};
+use alloc::vec::Vec;
+use bitcoin::{
+    secp256k1::{self, Secp256k1},
+    OutPoint, Script, TxOut,
+};
 use core::{fmt::Debug, ops::Deref};
 
 use super::DerivationAdditions;
@@ -50,7 +54,10 @@ const DERIVED_KEY_COUNT: u32 = 1 << 31;
 #[derive(Clone, Debug)]
 pub struct KeychainTxOutIndex<K> {
     inner: SpkTxOutIndex<(K, u32)>,
-    keychains: BTreeMap<K, Descriptor<DescriptorPublicKey>>,
+    // value: (descriptor, stored-index-count)
+    keychains: BTreeMap<K, (Descriptor<DescriptorPublicKey>, u32)>,
+    // lookahead settings for each keychain
+    lookahead: BTreeMap<K, u32>,
 }
 
 impl<K> Default for KeychainTxOutIndex<K> {
@@ -58,6 +65,7 @@ impl<K> Default for KeychainTxOutIndex<K> {
         Self {
             inner: SpkTxOutIndex::default(),
             keychains: BTreeMap::default(),
+            lookahead: BTreeMap::default(),
         }
     }
 }
@@ -98,8 +106,8 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         &self.inner
     }
 
-    /// Get the internal map of keychains to their descriptors.
-    pub fn keychains(&self) -> &BTreeMap<K, Descriptor<DescriptorPublicKey>> {
+    /// Get the internal map of keychains to their descriptors and the stored index count.
+    pub fn keychains(&self) -> &BTreeMap<K, (Descriptor<DescriptorPublicKey>, u32)> {
         &self.keychains
     }
 
@@ -109,7 +117,52 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     /// and the txout index will discover transaction outputs with those script pubkeys.
     pub fn add_keychain(&mut self, keychain: K, descriptor: Descriptor<DescriptorPublicKey>) {
         // TODO: panic if already different descriptor at that keychain
-        self.keychains.insert(keychain, descriptor);
+        self.keychains.insert(keychain, (descriptor, 0));
+    }
+
+    /// Convenience method of [`set_lookahead`] for all keychains.
+    ///
+    /// [`set_lookahead`]: Self::set_lookahead
+    pub fn set_all_lookaheads(&mut self, lookahead: u32) {
+        let secp = Secp256k1::verification_only();
+        for keychain in &self.keychains.keys().cloned().collect::<Vec<_>>() {
+            self.lookahead.insert(keychain.clone(), lookahead);
+            self.replenish_lookahead(&secp, &keychain);
+        }
+    }
+
+    /// Set the lookahead count for `keychain`. This is the number of scripts to cache ahead
+    /// of the last derived script index.
+    ///
+    /// ## Panics
+    ///
+    /// This will panic if `keychain` does not exist.
+    pub fn set_lookahead(&mut self, keychain: &K, lookahead: u32) {
+        self.lookahead.insert(keychain.clone(), lookahead);
+        self.replenish_lookahead(&Secp256k1::verification_only(), keychain);
+    }
+
+    fn replenish_lookahead<C>(&mut self, secp: &Secp256k1<C>, keychain: &K)
+    where
+        C: secp256k1::Verification,
+    {
+        let (descriptor, index_count) = self.keychains.get(keychain).expect("keychain must exist");
+        let index_count = *index_count;
+        let lookahead = self.lookahead.get(keychain).map_or(0, |v| *v);
+
+        for index in index_count..index_count + lookahead {
+            let spk = descriptor
+                .at_derivation_index(index)
+                .derived_descriptor(secp)
+                .map(|desc| desc.script_pubkey());
+
+            match spk {
+                Ok(spk) => self
+                    .inner
+                    .insert_script_pubkey((keychain.clone(), index), spk),
+                Err(_) => break,
+            };
+        }
     }
 
     /// Generates script pubkey iterators for every `keychain`. The iterators iterate over all
@@ -119,7 +172,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     ) -> BTreeMap<K, impl Iterator<Item = (u32, Script)> + Clone> {
         self.keychains
             .iter()
-            .map(|(keychain, descriptor)| {
+            .map(|(keychain, (descriptor, _))| {
                 (
                     keychain.clone(),
                     descriptor_into_script_iter(descriptor.clone()),
@@ -136,8 +189,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
     ) -> Option<impl Iterator<Item = (u32, Script)> + Clone> {
         self.keychains
             .get(keychain)
-            .cloned()
-            .map(descriptor_into_script_iter)
+            .map(|(descriptor, _)| descriptor_into_script_iter(descriptor.clone()))
     }
 
     /// Iterates over the script pubkeys derived and stored by this index of all keychains.
@@ -180,6 +232,7 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             .keychains
             .get(keychain)
             .expect(&format!("keychain {:?} does not exist", keychain))
+            .0
             .has_wildcard();
 
         match self.derivation_index(keychain) {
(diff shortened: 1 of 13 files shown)


bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -83,23 +91,37 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
     }
 
     pub fn inner(&self) -> &SpkTxOutIndex<(K, u32)> {
         &self.inner
     }
 
-    /// Get the internal map of keychains to their descriptors.
-    pub fn keychains(&self) -> &BTreeMap<K, Descriptor<DescriptorPublicKey>> {
+    /// Get the internal map of keychains to their descriptors and the stored index count.
+    pub fn keychains(&self) -> &BTreeMap<K, (Descriptor<DescriptorPublicKey>, u32)> {
```
LLFourn, line 124: I don't think we should change this API. Why is stored index count here? 

I recommend storing the current derivation index in another `BTreeMap<K, u32>`.

evanlinjin: I was actually thinking this just now! Will change soon.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -83,23 +91,37 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-        self.inner.scan_txout(op, &txout);
+    pub fn scan_txout(&mut self, op: OutPoint, txout: &TxOut) -> DerivationAdditions<K> {
+        let secp = Secp256k1::verification_only();
+        if let Some((keychain, index)) = self.inner.scan_txout(op, txout).cloned() {
+            let (_, index_count) = self
+                .keychains
+                .get_mut(&keychain)
+                .expect("keychain must exist");
+            if index >= *index_count {
+                *index_count = index + 1;
```
LLFourn, line 111: Would prefer to store current derivation index to avoid this `+1`.

bdk_chain/src/spk_txout_index.rs
```diff
@@ -64,18 +64,28 @@ impl<I: Clone + Ord> SpkTxOutIndex<I> {
     /// See [`ForEachTxout`] for the types that support this.
     ///
     /// [`ForEachTxout`]: crate::ForEachTxout
-    pub fn scan(&mut self, txouts: &impl ForEachTxout) {
-        txouts.for_each_txout(&mut |(op, txout)| self.scan_txout(op, txout))
+    pub fn scan(&mut self, txouts: &impl ForEachTxout) -> BTreeSet<I> {
```
LLFourn, line 67: `BTreeSet<&I>` to be consistent with rest of API.

bdk_chain/tests/test_keychain_txout_index.rs
```diff
@@ -34,52 +33,152 @@ fn init_txout_index() -> (
...
+    let external_index = txout_index
+        .derivation_index(&TestKeychain::External)
+        .expect("already derived");
+    let internal_index = txout_index
+        .derivation_index(&TestKeychain::Internal)
+        .expect("already derived");
+    assert_eq!(external_index, 19);
+    assert_eq!(internal_index, 24);
+
+    // lol
```
LLFourn, line 143: this is no laughing matter.

evanlinjin: I can imagine.

bdk_chain/src/spk_txout_index.rs
evanlinjin: @LLFourn I tried, but lifetime issues because we can't take reference out of closure apparently?

LLFourn: I also can't fix this. Maybe remove the return value here?

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Done

bdk_chain/src/spk_txout_index.rs
LLFourn: Yeah this is good just make a small internal function that is called by both functions?

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -107,16 +136,80 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+    ///
+    /// This will panic if a different `descriptor` is introduced to the same `keychain`.
     pub fn add_keychain(&mut self, keychain: K, descriptor: Descriptor<DescriptorPublicKey>) {
-        // TODO: panic if already different descriptor at that keychain
-        self.keychains.insert(keychain, descriptor);
+        let old_descriptor = &*self.keychains.entry(keychain).or_insert(descriptor.clone());
+        assert_eq!(
+            &descriptor, old_descriptor,
+            "keychain already contains a different keychain"
+        );
```
LLFourn, line 148: replenish lookahead?

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -82,23 +104,30 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
     ///
     /// If it matches the index will store and index it.
-    pub fn scan_txout(&mut self, op: OutPoint, txout: &TxOut) {
-        self.inner.scan_txout(op, &txout);
+    pub fn scan_txout(&mut self, op: OutPoint, txout: &TxOut) -> DerivationAdditions<K> {
+        let mut additions = DerivationAdditions::default();
+        if let Some((keychain, index)) = self.inner.scan_txout(op, txout).cloned() {
+            additions.append(self.store_up_to(&keychain, index));
+        }
+        additions
```
LLFourn, line 122: nit: clearer
```suggestion
pub fn scan_txout(&mut self, op: OutPoint, txout: &TxOut) -> DerivationAdditions<K> {
        match self.inner.scan_txout(op, txout).cloned() {
            Some((keychain, index)) => self.store_up_to(&keychain, index),
            None => Default::default(),
        }
```

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -107,16 +136,80 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+            };
+        }
     }
 
     /// Generates script pubkey iterators for every `keychain`. The iterators iterate over all
     /// derivable scripts.
-    pub fn scripts_of_all_keychains(
-        &self,
-    ) -> BTreeMap<K, impl Iterator<Item = (u32, Script)> + Clone> {
+    pub fn all_keychain_scripts(&self) -> BTreeMap<K, impl Iterator<Item = (u32, Script)> + Clone> {
```
LLFourn, line 212: Let's call the `script_pubkeys` not "scripts".

LLFourn (commented):
Left some nit comments. I'd like to get rid of "store" altogether here. We are storing those in the lookahead too. What about "revealed" e.g. "last_revealed_indices".

Along this line of thought can we remove `store_up_to` and `store_all_up_to` and replace them with `reveal(keychain, index)` and `reveal_all(btree_map_of_indices)` where they actually return the revealed script pubkeys.

So the notion is of spks I have revealed and those I have not.  Naming suggestions:

- `next_unused` stays as `next_unused`.
- `derive_new` becomes `reveal_next`.
- `DerivationAdditions` becomes `RevelationAdditions`.
- `stored_scripts` becomes `revealed_script_pubkeys`

wdyt?

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: This is to ensure caller does not store a different descriptor under the same keychain.

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Done

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Done.

evanlinjin:
@LLFourn 

* I had another go at renaming (I think it is a lot better now).
* `reveal..` methods now returned iterators of things they have newly revealed.
* I repurposed `descriptor_into_script_iter` into `range_descriptor_spks` where you can specify a range of indices (and the descriptor is wrapped in Cow to either be owned (for external use) or borrowed (for internal use)). This way it can be used efficiently within `reveal..` methods and `replenish_lookahead` while enforcing checks (such as non-wildcard descriptors should only have spk at 0) for them. We want to use this as a wrapper because miniscript library allows deriving non-wildcard descriptors above 0 (which returns the same script for every index) and this resulted in a bug).
* Updated tests and docs further.

bdk_chain/src/keychain/keychain_txout_index.rs
LLFourn: No I mean you need to replenish lookahead here :facepalm:

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -195,92 +284,130 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if `keychain` does not exist.
+    pub fn reveal_to_target(
+        &mut self,
+        keychain: &K,
+        target_index: u32,
+    ) -> (
+        Option<impl Iterator<Item = (u32, Script)>>,
```
LLFourn, line 340: What is the semantics of `None` here? If it just means the iterator would be empty just return an empty iterator.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -340,42 +467,10 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-        match up_to {
-            Some(up_to) => self.store_up_to(keychain, up_to),
-            None => Default::default(),
-        }
-    }
-
     /// Iterates over all unused script pubkeys for a `keychain` that have been stored in the index.
-    pub fn keychain_unused(&self, keychain: &K) -> impl DoubleEndedIterator<Item = (u32, &Script)> {
-        let range = (keychain.clone(), u32::MIN)..(keychain.clone(), u32::MAX);
+    pub fn unused_spks(&self, keychain: &K) -> impl DoubleEndedIterator<Item = (u32, &Script)> {
```
LLFourn, line 471: This new name is worse because it is inconsistent. The `inner` which we deref to has a method name `unused` which returns unused spks and now we have a method that is named `unused_spks` which actually gets unused by keychain. `unused_for_keychain` might be better.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -107,82 +134,144 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+        }
     }
 
     /// Generates script pubkey iterators for every `keychain`. The iterators iterate over all
-    /// derivable scripts.
-    pub fn scripts_of_all_keychains(
-        &self,
-    ) -> BTreeMap<K, impl Iterator<Item = (u32, Script)> + Clone> {
+    /// derivable script pubkeys.
+    pub fn keychain_spks_of_all(&self) -> BTreeMap<K, impl Iterator<Item = (u32, Script)> + Clone> {
```
LLFourn, line 202: s/keychain_spks_of_all/spks_of_all_keychains/g

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -107,82 +134,144 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
-    pub fn scripts_of_keychain(
-        &self,
-        keychain: &K,
-    ) -> Option<impl Iterator<Item = (u32, Script)> + Clone> {
-        self.keychains
+    ///
+    /// # Panics
+    ///
+    /// This will panic if `keychain` does not exist.
+    pub fn keychain_spks(&self, keychain: &K) -> impl Iterator<Item = (u32, Script)> + Clone {
```
LLFourn, line 220: same here. The old name was better but just replace scripts with spks.

bdk_chain/src/tx_data_traits.rs
```diff
@@ -7,14 +7,14 @@ use bitcoin::{Block, OutPoint, Transaction, TxOut};
...
+    fn for_each_txout(&self, f: impl FnMut((OutPoint, &TxOut)));
 }
 
-impl ForEachTxout for Block {
-    fn for_each_txout(&self, f: &mut impl FnMut((OutPoint, &TxOut))) {
+impl ForEachTxOut for Block {
+    fn for_each_txout(&self, mut f: impl FnMut((OutPoint, &TxOut))) {
         for tx in self.txdata.iter() {
-            tx.for_each_txout(f)
+            tx.for_each_txout(&mut f)
```
LLFourn, line 17: TIL `FnMut` is implemented for `&mut F where F: FnMut`.

LLFourn (commented):
This looks great. There are a few naming regressions. Also can we make it consistent over the whole code base whether we are using `spk` or `script_pubkey`. Choose one and make it consistent everywhere. Thanks.

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Fixed in 7e664feabc3c37919bb774d192a8864c1aaa7843

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Fixed in d2b0d214e6c7819aa904237c967c46bdaf66ab5e

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Fixed in d2b0d214e6c7819aa904237c967c46bdaf66ab5e

bdk_chain/src/keychain/keychain_txout_index.rs
evanlinjin: Fixed in d2b0d214e6c7819aa904237c967c46bdaf66ab5e

bdk_chain/src/tx_data_traits.rs
evanlinjin: I changed the trait to take in a closure without a reference... :/

PR #170, opened 2023-02-06
[test] RPC wallet syncing with BitcoinD

rajarshimaitra:
This is an initial draft of RPC wallet syncing tests using `electrsd::BitcoinD`. 

The general approach of the tests are
 - Create a `BitcoinD` instance in rgetest
 - Generate and fill blocks up to 101.
 - Use it to sync a `tracker` and a `db`.

The API isn't final, and I did it as a rough sketch. So no documentation is included. It uses the `RpcClient` sync code and the `main.rs` transaction fetching code to do the full syncing. 

Three test cases have been added so far; among them, 2 are failing.

 - `basic_sync_test` : Successfully syncing the chain and getting confirmed balance.
 - `basic_reorg()` (failing): After reorg balance changes to `unconfirmed_pending` and doesn't get confirmed even after generating new blocks. 
  - `basic_send_test()` (failing): `create_tx()` failing at `failed to obtain change plan`.
  
I am not fully sure why these tests are failing. But at least this provides us a debug ground to scour further. I will spend some time next to see if I can identify the reasons for the failure.

Any review comment or ideas on the test failure is welcome.


rajarshimaitra:
The latest update does the following:
 - Fix the failing tests.
 - Adds docs and comments.
 - Merge all the RPC and tests commits together.
 - Overrides #79 .

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,314 @@
...
+impl RpcClient {
+    /// Initialize [`RpcClient`] from a [`RpcConfig`] and [`KeychainTxOutIndex`].
+    /// This creates a wallet inside bitcoin core, with a deterministic name
+    /// derived from the hash of descriptors within the [`KeychainTxOutIndex`].
+    /// Then imports all the spks from the index into the newly created core wallet.
+    pub fn init_for_tracker<K: Debug + Clone + Ord>(
+        config: &RpcConfig,
+        txout_index: &KeychainTxOutIndex<K>,
+    ) -> Result<Self, RpcError> {
+        let wallet_name = derive_name_from_keychainindex(txout_index);
```
LLFourn, line 148: Why not let the caller pass in the wallet name and explain to them how to choose it? You may want to organize your wallets by a human friendly name.

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+impl RpcClient {
+    /// Initialize [`RpcClient`] from a [`RpcConfig`] and [`KeychainTxOutIndex`].
+    /// This creates a wallet inside bitcoin core, with a deterministic name
+    /// derived from the hash of descriptors within the [`KeychainTxOutIndex`].
+    /// Then imports all the spks from the index into the newly created core wallet.
+    pub fn init_for_tracker<K: Debug + Clone + Ord>(
+        config: &RpcConfig,
+        txout_index: &KeychainTxOutIndex<K>,
```
LLFourn, line 146: Why not just pass in an `impl Iterator<Item=Script>` or something simple like that.

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,314 @@
...
+    ) -> Result<Self, RpcError> {
+        let wallet_name = derive_name_from_keychainindex(txout_index);
+        let wallet_url = format!("{}/wallet/{}", config.url, wallet_name);
+        let client = Client::new(wallet_url.as_str(), config.auth.clone().into())?;
```
LLFourn, line 150: Please make the caller create the client and then call methods on it. That way we don't need to have any config defined in this library.

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,314 @@
...
+        let network = match blockchain_info.chain.as_str() {
+            "main" => Network::Bitcoin,
+            "test" => Network::Testnet,
+            "regtest" => Network::Regtest,
+            "signet" => Network::Signet,
+            _ => return Err(RpcError::General("Invalid Network string".to_string())),
+        };
+        if network != config.network {
+            return Err(RpcError::General("Invalid Network".to_string()));
+        }
```
LLFourn, line 191: Do we need to do this? I would rather leave this for the caller to check that they are on the network they think they are on.

bdk_rpc_wallet_example/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[dependencies]
+# BDK Core
+bdk_chain = { path = "../bdk_chain", features = ["serde"] }
+bdk_cli = { path = "../bdk_cli_lib"}
+
+# Electrum
+bitcoincore-rpc = "0.16"
```
LLFourn, line 13: This is not electrum :P

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,314 @@
...
+            .map(|i| (i.blocks, i.best_block_hash))?)
+    }
+
+    /// Scan for a keychain tracker, and create an initial [`SparseChain`] candidate update.
+    /// This update will only contain [`Txid`]s in SparseChain, and no actual transaction data.
+    ///
+    /// User needs to fetch the required transaction data and update the [`SparseChain`] before applying it.
+    pub fn wallet_scan(
+        &self,
+        local_chain: &BTreeMap<u32, BlockHash>,
```
LLFourn, line 218: shouldn't we pass in a wallet name here so we don't scan every wallet?

bdk_rpc_wallet_example/src/rpc.rs
LLFourn: Ok I looked at the docs -- bitcoin core rpc wallet seems to be insane in that you can only have one wallet loaded at a time.

Can we just let the caller handle this wallet creation thing..? i.e. handle it in `main.rs`.

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,314 @@
...
+    ) -> Result<SparseChain, RpcError> {
```
LLFourn, line 219: The `SparseChain` here should at least be `SparseChain<ConfirmationTime>` since rpc returns that info I think.

bdk_rpc_wallet_example/src/rpc.rs
```diff
@@ -0,0 +1,314 @@
...
+        let page_size = 1000; // Core has 1000 page size limit
+        let mut page = 0;
+
+        let _ = self.rescan_blockchain(Some(last_common_height as usize), None);
+
+        let mut txids_to_update = Vec::new();
+
+        loop {
+            let list_tx_result = self
+                .list_transactions(None, Some(page_size), Some(page * page_size), Some(true))?
```
LLFourn, line 262: wasn't the plan to use the label arugment (the first one) to differentiate between wallets (I was getting confused between this and wallet name).

LLFourn (changes requested):
Thanks @rajarshimaitra. I think the design works. I suggest trimming down the scope of `rpc.rs` to only `wallet_scan`.  Everything else seems like pure example code that can go in `main.rs`. The user can be left to make sure that all the spks they are interested in are loaded into the wallet. This allows them to use descriptors also. If we want to get more opinionated about this we can do it later.

The tests are not testing `wallet_scan`. I think they should **only** be testing the sparse chain output from `wallet_scan`.

I think we should make use of the label feature in bitcoin core rpc so `wallet_scan` takes a `label` to scan over (perhaps optional). In the example we should somehow expose the label.

Maybe call `wallet_scan` `fetch_sparse_chain`.

rajarshimaitra:
Thanks @LLFourn for the review.. Updated as suggested.

 - Helper functions moved to `main.rs`.
 - `wallet_sync` moved to `lib.rs`.
 - Couldn't remove the struct `RpcClient` as `bdk_cli::handle_commands` expect a generic trait bound `Broadcast` and we can't impl foreign traits on foreign types.
  - Blockchain tests only testing for sparsechain structure.
  - We can't do create transaction test if we aren't fully syncing the tracker, which is the case currently as we are only getting the `update_sparsechain` from `wallet_sync`.

LLFourn:
Thanks @rajarshimaitra.

>     * Helper functions moved to `main.rs`.
> 
>     * `wallet_sync` moved to `lib.rs`.
> 
>     * Couldn't remove the struct `RpcClient` as `bdk_cli::handle_commands` expect a generic trait bound `Broadcast` and we can't impl foreign traits on foreign types.

We could fix this by getting rid of the `Broadcast` trait and just passing in a closure to the cli lib's main handler that takes a tx and broadcasts it.

> 
>     * Blockchain tests only testing for sparsechain structure.
> 
>     * We can't do create transaction test if we aren't fully syncing the tracker, which is the case currently as we are only getting the `update_sparsechain` from `wallet_sync`.

This is good because this library doesn't provide create tx functionality.

rajarshimaitra:
> We could fix this by getting rid of the Broadcast trait and just passing in a closure to the cli lib's main handler that takes a tx and broadcasts it. See https://github.com/LLFourn/bdk_core_staging/pull/198

That totally makes sense.. Will update accordingly.. Should I just build this over #198 or port the change?

PR #172, opened 2023-02-07
`TxGraph::full_transactions` should return `&T` to be consistent

evanlinjin:


LLFourn:
oops I did this myself just now.

PR #175, opened 2023-02-09
s/DERIVED_KEY_COUNT/BIP32_MAX_INDEX/

LLFourn:
This constant is not a "count" but an upper bound which we should conventionally call a "max".

I made it public since it might be useful to refer to it in docs later.


evanlinjin (approved):
ACK cb45797ad345bdecf9c3f37fa35c4b6b97b99932

PR #182, opened 2023-02-11
Add `revealed_and_lookahead_spks_of_keychain`

danielabrozzoni:
Sometimes we want to know the scripts revealed and the one present in the lookahead (for example, in the CBF example)

### DIFF
--- a/bdk_chain/src/keychain/keychain_txout_index.rs
+++ b/bdk_chain/src/keychain/keychain_txout_index.rs
@@ -253,6 +253,38 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
             .map(|((_, derivation_index), spk)| (*derivation_index, spk))
     }
 
+    /// Convenience method to get [`revealed_and_lookahead_spks_of_keychain`]
+    /// of all keychains.
+    ///
+    /// [`revealed_and_lookahead_spks_of_keychain`]: Self::revealed_and_lookahead_spks_of_keychain
+    pub fn revealed_and_lookahead_spks_of_all_keychains(
+        &self,
+    ) -> BTreeMap<K, impl Iterator<Item = (u32, &Script)> + Clone> {
+        self.keychains
+            .keys()
+            .map(|keychain| {
+                (
+                    keychain.clone(),
+                    self.revealed_and_lookahead_spks_of_keychain(keychain),
+                )
+            })
+            .collect()
+    }
+
+    /// Iterates over the script pubkeys revealed and the ones contained in
+    /// the lookahead under `keychain`.
+    pub fn revealed_and_lookahead_spks_of_keychain(
+        &self,
+        keychain: &K,
+    ) -> impl DoubleEndedIterator<Item = (u32, &Script)> + Clone {
+        let lookahead = self.lookahead.get(keychain).unwrap_or(&0);
+        let next_index = self.last_revealed.get(keychain).map_or(0, |v| *v + 1) + lookahead;
+        self.inner
+            .all_spks()
+            .range((keychain.clone(), u32::MIN)..(keychain.clone(), next_index))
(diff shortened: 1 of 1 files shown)


danielabrozzoni:
Oh wait, I just realized you can also do `txout_index.inner().all_spks()`, maybe it's better to document this instead of adding yet another method?

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -253,6 +253,38 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+                    keychain.clone(),
+                    self.revealed_and_lookahead_spks_of_keychain(keychain),
+                )
+            })
+            .collect()
+    }
+
+    /// Iterates over the script pubkeys revealed and the ones contained in
+    /// the lookahead under `keychain`.
+    pub fn revealed_and_lookahead_spks_of_keychain(
```
rajarshimaitra, line 276: we could also name it `stored_spks()` or something like that. the names are now getting a bit too verbose, 😅

We also should document somewhere the difference between all three kinds of spk lookups.

bdk_chain/src/keychain/keychain_txout_index.rs
```diff
@@ -253,6 +253,38 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
...
+        &self,
+        keychain: &K,
+    ) -> impl DoubleEndedIterator<Item = (u32, &Script)> + Clone {
+        let lookahead = self.lookahead.get(keychain).unwrap_or(&0);
+        let next_index = self.last_revealed.get(keychain).map_or(0, |v| *v + 1) + lookahead;
+        self.inner
+            .all_spks()
+            .range((keychain.clone(), u32::MIN)..(keychain.clone(), next_index))
+            .map(|((_, derivation_index), spk)| (*derivation_index, spk))
+    }
```
rajarshimaitra, line 286: I might be wrong, but maybe it is useful to error in case of non existing keychain than give back an empty? 

We don't enforce this behavior consistently yet. In `spks_of_keychain` it is an `expect` so the call panics for wrong keychain. While in `revealed_spks_of_keychain` it doesn't and gives empty iterator. 

I am not sure whats the right answer here, but to me it seems better to just panic early in this particluar case because it's clearly an user error?

rajarshimaitra (commented):
Thanks @danielabrozzoni  Code review ACK. dbfb861655a15fe7d4334a5645d7692a21e6f9ab

This gives what I was looking for.. 
Below are few comments and questions.

rajarshimaitra:
> Oh wait, I just realized you can also do txout_index.inner().all_spks(), maybe it's better to document this instead of adding yet another method?

I think it's still valuable to have the `revealed_and_lookahead_spks_of_keychain` function, which is just filtering for a specific keychain. It makes sense to have the API to get only stored scipts of a keychain. Or else the user need to figure out our indexing system and do the filtering themselves. But for `all keychains` we can essentially skip the extra API and document somewhere to just call ` txout_index.inner().all_spks()`.

rajarshimaitra:
> Oh wait, I just realized you can also do txout_index.inner().all_spks(), maybe it's better to document this instead of adding yet another method?

I think it's still valuable to have the `revealed_and_lookahead_spks_of_keychain` function, which is just filtering for a specific keychain. It makes sense to have the API to get only stored scipts of a keychain. Or else the user need to figure out our indexing system and do the filtering themselves. But for `all keychains` we can essentially skip the extra API and document somewhere to just call ` txout_index.inner().all_spks()`.

LLFourn:
> Oh wait, I just realized you can also do `txout_index.inner().all_spks()`, maybe it's better to document this instead of adding yet another method?

I think this is the right way forward. It needs to be clear that you can always get the inner txout index which has all the spks that the thing actually scans for.

PR #191, opened 2023-02-15
Electrum update with outpoints and txids

### LINKED ISSUE #84: Esplora and electrum should allow producing an update based on outputs rather than script pubkey
Right now the only way to find out if an output is spent is to query the APIs for a script pubkey. What if I only have an outpoint that I am interested in or I just want to check if a transaction is confirmed.

In addition to the iterator of script pubkeys we currently pass in we should allow iterators for:

1. `OutPoint` which would return the tx with the outpoint on it (if it can find it) and any tx spending from it
2. `Txid` which would return the transaction with this txid in the update if it exists.

evanlinjin:
Related to #84 

* Split `bdk_electrum` (library) out of `bdk_electrum_example`.
* Changed `ElectrumClient` logic to allowing syncing with arbitrary spks, outpoints, txids, full txs.
* Remove `ElectrumError::Reorg` variant and implement retry logic to mitigate reorgs.
* Introduced a bunch of convenience methods.
* Docs.


bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,766 @@
...
+    }
+}
+
+/// Parameters for [`ElectrumClient::scan`].
+pub struct ScanParams<K, S: IntoIterator<Item = (u32, Script)>> {
+    /// Indexed script pubkeys of each keychain to scan transaction histories for.
+    pub keychain_spks: BTreeMap<K, S>,
+    /// Arbitary script pubkeys to scan transaction histories for, which are not associated to a
+    /// keychain.
+    pub arbitary_spks: Vec<Script>,
```
LLFourn, line 629: We need this to be an iterator too so we know what progress is being made. It should be called `misc_spks` rather than arbitrary (they are not arbitrary in most senses). 

I would still prefer not to have this and force the caller to put this into a keychain. You can have a convenience method which creates a `ScanParams` from an iterator of `Script`.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,766 @@
...
+                    sparse_chain::InsertTxError::TxMovedUnexpectedly { .. } => {
+                        return Err(InternalError::Reorg);
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    fn populate_with_txs<'a>(
```
LLFourn, line 319: many methods here are only called in one place. Please inline them to their call location so the code is easier to reason about.

LLFourn (changes requested):
This PR implementation looks good and could be much better if its scope was severely reduced. You over egged this one a bit. This should only define `bdk_electrum` with the most minimal API surface area possible to meet our needs. Here's what I would like:

1. Remove all convenience methods in `bdk_chain`. There should be new functions in `bdk_chain`. You can opportunistically add some convenience  stuff for dealing with electurm data in `bdk_electurm`. 
2. `bdk_electrum` should not define its own error types and should wrap the client. It should only use extension traits with no custom errors.

Don't worry that the user has to call a few things and put a few things together to produce an update properly. Making it too simple is just as bad as making it too complicated at this stage. We want to show the guts of `bdk_chain` to people and for people to understand what a `ChainGraph`, `SparseChain` etc is when they use it rather than hiding these details from them.

bdk_electrum/src/lib.rs
evanlinjin: Done in bbff340

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,766 @@
...
+}
+
+/// Error type returned by [`ElectrumClient`].
+#[derive(Debug)]
+pub enum ElectrumError {
+    /// Error of the internal [`electrum_client::Client`].
+    Client(electrum_client::Error),
+    /// Error that occurs when the provided outpoint (via [`ScanParams`]) does not exist on the
+    /// transaction (`vout` is too high).
+    InvalidOutPoint(OutPoint),
```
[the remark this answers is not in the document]
evanlinjin: Fixed in 94f6409c3938b6a5056842fe8474c6a187ea974d

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,766 @@
...
+//! # #[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Debug)]
+//! # enum MyKeychain {
+//! #     External,
+//! #     Internal,
+//! # }
+//! // Create our electrum client.
+//! let client = ElectrumClient::<TxHeight>::from_config(url, config).expect("must connect");
+//!
+//! // Initialize our tracker (we want to update this).
+//! let mut tracker = KeychainTracker::<MyKeychain, TxHeight, Transaction>::default();
```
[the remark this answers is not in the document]
evanlinjin: Changed in 6fb851ca93509fdc6907d30b7a72687a8ff0b51a

bdk_chain/src/keychain/keychain_tracker.rs
```diff
@@ -59,6 +67,62 @@ where
...
+    pub fn set_lookahead_to_targets(&mut self, targets: BTreeMap<K, u32>) {
+        self.txout_index.lookahead_to_target_multi(targets)
+    }
+
+    /// Convenience method to call [`find_missing_txids`] of the internal [`ChainGraph`].
+    ///
+    /// [`find_missing_txids`]: ChainGraph::find_missing_txids
+    pub fn find_missing_txids<U: AsRef<SparseChain<P>>>(&self, chain_update: U) -> Vec<Txid> {
+        self.chain_graph().find_missing_txids(chain_update)
+    }
```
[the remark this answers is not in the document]
evanlinjin: Removed in b9f26a93eab5016bc83430fa7c62e125e9ac16a0

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,766 @@
...
+impl<K: Ord + Clone + Debug, P: ChainPosition> ScanUpdate<K, P> {
+    /// Apply the [`ScanUpdate`] to the `tracker`.
+    ///
+    /// This will fail if there are missing full transactions not provided via `new_txs`.
+    pub fn apply(
+        self,
+        new_txs: Vec<Transaction>,
+        tracker: &mut KeychainTracker<K, P>,
+    ) -> Result<KeychainChangeSet<K, P, Transaction>, InflateAndUpdateError<P>> {
+        tracker.set_lookahead_to_targets(self.last_active_indices);
```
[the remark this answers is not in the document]
evanlinjin: All done in b9f26a93eab5016bc83430fa7c62e125e9ac16a0

bdk_electrum/src/lib.rs
evanlinjin: Done in c1a0443999af9842a16cf9f92c26250e79189f0d

bdk_electrum_example/src/main.rs
```diff
@@ -33,6 +36,10 @@ enum ElectrumCommands {
         #[clap(flatten)]
         scan_option: ScanOption,
     },
+    /// Scan unspent outpoints for spends or changes to confirmation status of residing tx
+    SyncUnspent,
```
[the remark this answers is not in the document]
evanlinjin: Fixed in 0c225b8a15aad174812cf250d10712cb40fff332

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,766 @@
...
+    sparse_chain::{self, ChainPosition, SparseChain},
+    BlockId, ConfirmationTime, TxHeight,
+};
+use electrum_client::{Client, ElectrumApi, GetHistoryRes};
+
+/// Represents a [`ChainPosition`] that can be created via [`ElectrumClient`].
+///
+/// This is used internally within [`ElectrumClient`] so updates can be created for any
+/// [`ChainPosition`] that implements [`ElectrumChainPosition`].
+pub trait ElectrumChainPosition: ChainPosition + Sized {
```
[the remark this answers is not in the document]
evanlinjin: Changed to use methods here: 34773166a852fe6ff2bf217f0f72261d2f7527ba

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+            txids: Default::default(),
+            txs: Default::default(),
+            outpoints: Default::default(),
+            stop_gap: 10,
+            batch_size: 10,
+        }
+    }
+}
+
+impl<K, S: IntoIterator<Item = (u32, Script)>> From<BTreeMap<K, S>> for ScanParams<K, S> {
```
LLFourn, line 624: Make a constructor for this I think. Not sure what `From` is getting us since it makes it harder for the user to discover it.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+            if !params.txs.is_empty() {
+                match self.populate_with_txs(&to_position, &mut update, params.txs.iter()) {
+                    Err(InternalError::Reorg) => continue,
+                    Err(InternalError::ElectrumError(e)) => return Err(e.into()),
+                    Ok(_) => {}
+                }
+            }
+
+            if !params.outpoints.is_empty() {
+                match self.populate_with_outpoints(
```
LLFourn, line 523: All of these functions are called only once and so inlining them would make it easier to understand what is happening.

Also I think that we can remove `populate_with_spks` too. This will allow for removing `InternalError` and all these repeated match statements completely. Consider that for each loop iteration you can create an iterator of all the "already scanned" spks and the new ones and just take batches from there.

This comment is only a suggestion. I care much more about the public API!

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+}
+
+/// Parameters for [`ElectrumClient::scan`].
+pub struct ScanParams<K, S: IntoIterator<Item = (u32, Script)>> {
+    /// Indexed script pubkeys of each keychain to scan transaction histories for.
+    pub keychain_spks: BTreeMap<K, S>,
+    /// Txids to scan for. The update will update the [`ChainPosition`]s for these.
+    pub txids: Vec<Txid>,
+    /// Full transactions to scan for. The update will update the [`ChainPosition`]s for these.
+    pub txs: Vec<Transaction>,
```
LLFourn, line 600: I see this is here because it's faster to look up if we already have the tx. Can we not though. It's a super weird API!

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+    /// Outpoints to scan for. The update will try include the transaction that spends this outpoint
+    /// alongside the transaction which contains this outpoint.
+    pub outpoints: Vec<OutPoint>,
```
LLFourn, line 603: Note: these should also be iterators since you want to know the progress on them also. This can be done in another PR.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+            last_active_indices: self.last_active_indices,
+        })?)
+    }
+}
+
+impl<P> ElectrumUpdate<(), P> {
+    /// Convert [`ElectrumUpdate<(), P>`] into [`ElectrumUpdate<K, P>`].
+    pub fn into_with_keychain<K>(self) -> ElectrumUpdate<K, P> {
+        ElectrumUpdate {
+            update: self.update,
```
LLFourn, line 737: can we rename this to `chain_update` (in both places!).

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+                    {
+                        continue;
+                    }
+                    has_spending = true;
+                    res_tx
+                };
+                full_txs.insert(res.tx_hash, res_tx);
+
+                let tx_height = match res.height {
+                    h if h <= 0 => {
```
LLFourn, line 234: height == 0 means unconfirmed?? can we leave a fat comment explaining that.

LLFourn (changes requested):
Just took another look. A few things left to cleanup:

1. Make getting the extra position info another call (don't try and get position info in the main scan function as this is slow and an unwieldy).
2. Put the public API into an extension trait on `electrum_client::Client`
3. Remove `ScanParamsWithoutKeychain` in favor of a constructor on `ScanParams<(), S>`.

One idea I had while looking at this is that maybe it's worth looking at how lightning is implemented in the electrum wallet itself. Surely they have to do lookups via outpoint etc for that?

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+    fn from(value: BTreeMap<K, S>) -> Self {
+        Self {
+            keychain_spks: value,
+            ..Default::default()
+        }
+    }
+}
+
+/// Parameters for [`ElectrumClient::scan_without_keychain`].
+pub struct ScanParamsWithoutKeychain<S: IntoIterator<Item = Script>> {
```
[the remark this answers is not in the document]
evanlinjin: The iterators have different types so we need to map it (which takes in a closure) so it doesn't work unless we use `dyn`

bdk_electrum/src/lib.rs
evanlinjin: That will be a lot of generics or trait objects.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,800 @@
...
+        Ok(tracker.determine_changeset(&KeychainScan {
+            update: tracker.chain_graph().inflate_update(self.update, new_txs)?,
+            last_active_indices: self.last_active_indices,
+        })?)
+    }
+}
+
+impl<P> ElectrumUpdate<(), P> {
+    /// Convert [`ElectrumUpdate<(), P>`] into [`ElectrumUpdate<K, P>`].
+    pub fn into_with_keychain<K>(self) -> ElectrumUpdate<K, P> {
```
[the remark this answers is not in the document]
evanlinjin: Since the `chain_update` field is already public, I will remove this.

bdk_electrum/src/lib.rs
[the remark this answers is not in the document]
LLFourn: Which iterators have different types? There's only one iterator here...

bdk_electrum/src/lib.rs
LLFourn: There's no cost to having more generics :)

bdk_electrum/src/lib.rs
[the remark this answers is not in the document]
evanlinjin: https://github.com/LLFourn/bdk_core_staging/blob/4987af65d1991ebdd628d5ee12b57630d603ee7d/bdk_electrum/src/lib.rs#L596

https://github.com/LLFourn/bdk_core_staging/blob/4987af65d1991ebdd628d5ee12b57630d603ee7d/bdk_electrum/src/lib.rs#L636

bdk_electrum/src/lib.rs
evanlinjin: Actually, this is not needed. Removed.

bdk_electrum/src/lib.rs
evanlinjin: Removed these structs as per private dm.

bdk_electrum/src/lib.rs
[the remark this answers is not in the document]
evanlinjin: struct removed.

bdk_electrum/src/lib.rs
evanlinjin: I think we can refactor in a separate PR for this. Although, I'm not yet sold on in-lining everything...

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+                    let h = h as u32;
+                    if h > tip.height {
+                        TxHeight::Unconfirmed
+                    } else {
+                        TxHeight::Confirmed(h)
+                    }
+                }
+            };
+
+            if let Err(failure) = update.insert_tx(txid, tx_height) {
```
danielabrozzoni, line 449: Can you double check this line? It seems to me that `txid` refers to `tx`, but `tx_height` might not, as it might refer to the tx spending

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+            if has_residing && has_spending {
+                break;
+            }
+
+            // skip if we have already added the tx to our `update`
+            if full_txs.contains_key(&res.tx_hash) {
+                if res.tx_hash == txid {
+                    has_residing = true;
+                } else {
+                    has_spending = true;
```
danielabrozzoni, line 407: Are you sure about this? You're saying that if `res` is already in `full_txs` then it must be the tx spending from the initial one. Are you sure this holds?

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+            .map(|r| r.height)
+        {
+            Some(h) if h > 0 => TxHeight::Confirmed(h as _),
+            Some(_) => TxHeight::Unconfirmed,
+            None => continue,
+        };
+
+        if tx_height.is_confirmed() && tx_height > TxHeight::Confirmed(tip.height) {
+            continue;
+        }
```
danielabrozzoni, line 500: Shouldn't you insert tx as unconfirmed in this case, as you do above, in populate_outpoints?

bdk_electrum_example/src/main.rs
```diff
@@ -23,13 +24,20 @@ enum ElectrumCommands {
     Sync {
         /// Scan all the unused addresses
         #[clap(long)]
```
danielabrozzoni, line 26: I would provide some defaults for these args

bdk_electrum_example/src/main.rs
```diff
@@ -89,137 +122,141 @@ fn main() -> anyhow::Result<()> {
...
-                }
-                let local_chain = tracker.chain().checkpoints().clone();
+            // Get a short lock on the tracker to get the spks we're interested in
+            let tracker = tracker.lock().unwrap();
 
-                (spks, local_chain)
-            };
+            if !(all_spks || unused_spks || unspent_spks || unspent_outpoints || unconfirmed_txs) {
+                unused_spks = true;
+                unspent_spks = true;
```
danielabrozzoni, line 164: I wonder if there's a better way to enforce that at least one of the args is true, this looks a bit weird

bdk_electrum_example/src/main.rs
```diff
@@ -89,137 +122,141 @@ fn main() -> anyhow::Result<()> {
...
+            }
 
-            // we scan the spks **without** a lock on the tracker
-            let new_sparsechain = client
-                .spk_txid_scan(spks, &local_chain, scan_option.batch_size)
-                .context("scanning the blockchain")?;
+            let mut spks: Box<dyn Iterator<Item = bdk_chain::bitcoin::Script>> =
+                Box::new(core::iter::empty());
+            if all_spks {
+                let all_spks = tracker
```
danielabrozzoni, line 173: Does this "all" mean "just the stored spks" or does it really mean **all** the 2^30something possible spks?

bdk_electrum_example/src/main.rs
```diff
@@ -89,137 +122,141 @@ fn main() -> anyhow::Result<()> {
...
-            .map(|&(_, txid)| txid)
-            .collect::<Vec<_>>()
-    };
+    let missing_txids = response.find_missing_txids(&*tracker.lock().unwrap());
 
     // fetch the missing full transactions **without** a lock on the tracker
     let new_txs = client
-        .batch_transaction_get(new_txids.iter())
+        .batch_transaction_get(missing_txids)
         .context("fetching full transactions")?;
```
danielabrozzoni, line 248: I'm wondering if it would make sense to have a `fetch_missing_txs` that basically merges `find_missing_txids` and the `batch_transaction_get`, instead of `find_missing_txids`.

I don't think anyone would use `find_missing_txids` if it's not for then fetching them, and having a method that does both operations might be clearer for the user

danielabrozzoni (commented):
Looks nice to me, I have a few questions. I definitely need to do another pass, I didn't look at it too deeply

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+        .expect("update must atleast have one checkpoint");
+
+    let mut full_txs = HashMap::new();
+    for outpoint in outpoints {
+        let txid = outpoint.txid;
+        let tx = client.transaction_get(&txid)?;
+        debug_assert_eq!(tx.txid(), txid);
+        let txout = match tx.output.get(outpoint.vout as usize) {
+            Some(txout) => txout,
+            None => continue,
```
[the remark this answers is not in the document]
evanlinjin: That was my intuition initially as well!

This is @LLFourn's argument against an extra error case: https://github.com/LLFourn/bdk_core_staging/pull/191#discussion_r1111409193, which I agree with now.

It seems this error case is not too "useful"? Wdyt?

bdk_electrum/src/lib.rs
evanlinjin: `update` is a `SparseChain`, so it's `txid` here!

bdk_electrum/src/lib.rs
evanlinjin: Good catch! I think you are correct.

bdk_electrum/src/lib.rs
danielabrozzoni: I'm not sure I understand :(

What I meant is: `txid` is referring to the initial `tx`, but `tx_height` refers to `res`, and `res` is one of the txs returned by `client.script_get_history(&txout.script_pubkey)?`. So I'm wondering if you're inserting a `txid` with a confirmation height that belongs to a different transaction

bdk_electrum/src/lib.rs
evanlinjin: Yes. Thanks again

bdk_electrum_example/src/main.rs
evanlinjin: All stored.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+            .into_iter()
+            .map(|(k, s)| {
+                let iter = s.into_iter();
+                (k, iter)
+            })
+            .collect::<BTreeMap<K, _>>();
+        let mut scanned_spks = BTreeMap::<(K, u32), (Script, bool)>::new();
+
+        let txids = txids.into_iter().collect::<Vec<_>>();
+        let outpoints = outpoints.into_iter().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
evanlinjin: There is a way, and this is the way we deal with scanning by spks.

So the `populate_with_..` methods needs to return a list of things it has scanned for.

The other option is to make these `IntoIterator` inputs `Clone` or `Copy`.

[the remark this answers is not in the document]
danielabrozzoni: 😂  it's ok, I think it's fine as-is

bdk_electrum/src/lib.rs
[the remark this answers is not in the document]
danielabrozzoni: I don't have a strong opinion, it just felt a bit weird, but I agree that it's not really useful to return an error

bdk_electrum/src/lib.rs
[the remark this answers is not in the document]
evanlinjin: Someone will fix it later 🤣

bdk_electrum/src/lib.rs
evanlinjin: `res_tx` should always be the same tx as that of `tx` at that point (so I hope).

Check lines from `412`:
```rust
            let res_tx = if res.tx_hash == txid {
                has_residing = true;
                tx.clone()
            } else {
                let res_tx = client.transaction_get(&res.tx_hash)?;
                if !res_tx
                    .input
                    .iter()
                    .any(|txin| txin.previous_output == outpoint)
                {
                    continue;
                }
                has_spending = true;
                res_tx
            };
```

danielabrozzoni: The else branch is returning the transaction spending from `tx` I think
```
            } else {
                let res_tx = client.transaction_get(&res.tx_hash)?;
                if !res_tx
                    .input
                    .iter()
                    .any(|txin| txin.previous_output == outpoint)
                {
                    continue;
                }
                has_spending = true;
                res_tx
            };
```

evanlinjin: Oh damn, good point I completely missed that thanks!

evanlinjin: Fixed in 97f5bec7266138567c49e6f8f343e86fe0227464

bdk_electrum/src/lib.rs
evanlinjin: Fixed in 97f5bec7266138567c49e6f8f343e86fe0227464

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,583 @@
+//! This crate is used for updating structures of the [`bdk_chain`] crate with data from electrum.
+//!
+//! The star of the show is the [`ElectrumExt::scan`] method, which scans for relevant blockchain
+//! data (via electrum) and outputs an [`ElectrumUpdate`].
+//!
+//! An [`ElectrumUpdate`] only includes `txid`s and no full transactions. The caller is responsible
+//! for obtaining full transactions before applying. This can be done with
+//! these steps:
+//!
+//! 1. Determine which full transactions are missing. The method [`find_missing_txids`] of
```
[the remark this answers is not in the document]
evanlinjin: How about `missing_full_txs`?

bdk_electrum_example/src/main.rs
evanlinjin: I think we should defer this.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,583 @@
...
+    ///
+    /// Refer to [crate-level documentation] for more.
+    ///
+    /// [crate-level documentation]: crate
+    fn scan<K: Ord + Clone>(
+        &self,
+        local_chain: &BTreeMap<u32, BlockHash>,
+        keychain_spks: BTreeMap<K, impl IntoIterator<Item = (u32, Script)>>,
+        txids: impl IntoIterator<Item = Txid>,
+        outpoints: impl IntoIterator<Item = OutPoint>,
```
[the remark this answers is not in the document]
evanlinjin: Updated in b8a4f3e30ecbba0fa28c2ea9436a9b5402ea2abc

bdk_electrum/src/lib.rs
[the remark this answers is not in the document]
evanlinjin: Changed to `missing_full_txs`.

bdk_electrum/src/lib.rs
evanlinjin: Fixed.

bdk_electrum_example/src/main.rs
evanlinjin: The defaults are just "false" for now, we need to do the weird logic below, let's defer this!

bdk_electrum_example/src/main.rs
evanlinjin: Let's defer this!

evanlinjin:
I've already released this! https://crates.io/crates/bdk_electrum

bdk_chain/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bdk_chain"
-version = "0.1.1"
+version = "0.2.0"
```
LLFourn, line 3: In general you shouldn't change the package version in this kind of PR.

Are you just doing this because you're making releases at the same time? I see that `bdk_chain` on crates.io is already at `0.3.0`.

bdk_chain/src/chain_graph.rs
```diff
@@ -39,6 +39,24 @@ impl<P, T> Default for ChainGraph<P, T> {
...
+    }
+}
+
+impl<P, T> AsRef<TxGraph<T>> for ChainGraph<P, T> {
+    fn as_ref(&self) -> &TxGraph<T> {
+        &self.graph
+    }
+}
+
+impl<'a, P, T> AsRef<ChainGraph<P, T>> for &'a ChainGraph<P, T> {
```
LLFourn, line 54: Ah so this is the PR its from. 
```suggestion
impl<P, T> AsRef<ChainGraph<P, T>> for ChainGraph<P, T> {
```
is what you want.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,608 @@
...
+                .as_ref()
+                .inflate_update(self.chain_update, new_txs)?,
+            last_active_indices: self.last_active_indices,
+        })
+    }
+}
+
+impl<K: Ord + Clone + Debug> ElectrumUpdate<K, TxHeight> {
+    /// Creates [`ElectrumUpdate<K, ConfirmationTime>`] from [`ElectrumUpdate<K, TxHeight>`].
+    pub fn new_confirmation_time_update(
```
LLFourn, line 269:
```suggestion
    pub fn into_confirmation_time_update(
```

And the receiver should be `self`.

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    Ok(update)
+}
+
+fn populate_with_outpoints(
+    client: &Client,
+    update: &mut SparseChain,
+    outpoints: &mut impl Iterator<Item = OutPoint>,
+) -> Result<HashMap<Txid, Transaction>, InternalError> {
```
LLFourn, line 380: worth putting a comment about why this is so damn awkward to do.

bdk_chain/Cargo.toml
evanlinjin: I had to release it for use with the other crates within the repo. Also, I was using the crate for the BDK PR as well (So I needed a new release with the persistence stuff).

bdk_chain/src/chain_graph.rs
evanlinjin: Does that work? Do you mean:
```rust
impl<P, T> AsRef<ChainGraph<P, T>> for &ChainGraph<P, T> { .. }
```

evanlinjin: Oh damn, it works

bdk_electrum/src/lib.rs
```diff
@@ -0,0 +1,608 @@
...
+                }
+                continue;
+            } else {
+                scanned_spks.insert(spk_index, (spk, true));
+                unused_spk_count = 0;
+            }
+
+            for tx in spk_history {
+                let tx_height = match tx.height {
+                    // Electrum server API specifies that height <= 0 is considered unconfirmed
```
LLFourn, line 575: So the genesis block tx is literally impossible to represent correctly :facepalm: ? It sounds like we have to special case the tx to check if it's the genesis tx so it doesn't show that this tx is always unconfirmed (any by extension all descendants are unconfirmed).

bdk_electrum/src/lib.rs
evanlinjin: But if it errors, the caller will loose "self". Would this be okay?

LLFourn (approved):
Minor nits. LGTM. Testing will have to come later.

bdk_electrum/src/lib.rs
evanlinjin: On it now.

bdk_electrum/src/lib.rs
LLFourn: Sure they can clone it if they want to keep it in this case.

evanlinjin: Agreed.

bdk_electrum/src/lib.rs
evanlinjin: Done

PR #192, opened 2023-02-15
Fix `KeychainTxOutIndex` lookahead logic

evanlinjin:
The lookahead logic tries to be as efficient as possible and avoids storing spks that are already stored. However, the original logic failed when the lookahead setting was changed more than once. This fixes it.

Additionally, the methods `lookahead_to_target` and `lookahead_to_target_multi` are added. These store unrevealed scripts without affecting lookahead settings.


danielabrozzoni (approved):
ACK 8f98288bfb46f14318d9c4f930ed9461ce6251db - I've been able to reproduce locally the failure, and the code looks good to me

PR #194, opened 2023-02-21
Add KeychainTxOutIndex::reserve_next_unused_spk #188

### LINKED ISSUE #188: Add KeychainTxOutIndex::reserve_next_unused_spk
Just like `next_unused_spk` but also marks it as used with `mark_used`. It's annoying to do this in two steps outside because of mutable borrow issues.

Should also have one for revealing and reserving at the same time.

Alternatively we could have returned owned `Script`s from all spk APIs. I liked returning references because it demonstrates what is actually stored in the index and what is being derived on the fly.

LagginTimes:
Fixes #188 
Second method will be done later this week.

### DIFF
--- a/bdk_chain/src/keychain/txout_index.rs
+++ b/bdk_chain/src/keychain/txout_index.rs
@@ -460,6 +460,30 @@ impl<K: Clone + Ord + Debug> KeychainTxOutIndex<K> {
         ((next_index, script), additions)
     }
 
+    /// Attempts to reveal the next script pubkey of the provided `keychain` and mark it as used.
+    ///
+    /// This is a convenience method that is equivalent to calling [`reveal_next_spk`] and
+    /// [`mark_used`].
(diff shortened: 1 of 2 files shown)


LLFourn:
LGTM! I think I still like this approach over the alternative which is to return an owned `Script` from `next_unused_spk`.

evanlinjin:
> LGTM! I think I still like this approach over the alternative which is to return an owned `Script` from `next_unused_spk`. 

Should we change `next_unused_spk` to return a referenced spk as well?

evanlinjin:
LGTM as well. I was talking to @LLFourn about this PR today and he suggested to add tests for these added methods. Let me know if you need some direction for this :)

evanlinjin (approved):
ACK d183b9dc8a113987bda36b9a294748aeb7799a4d

LGTM!

PR #196, opened 2023-02-21
Create `bdk_esplora` package

### LINKED ISSUE #84: Esplora and electrum should allow producing an update based on outputs rather than script pubkey

evanlinjin:
Fixes #84

Note that this is based on the work done in #191


bdk_esplora/src/lib.rs
```diff
@@ -4,13 +4,14 @@ use bdk_chain::{
     keychain::KeychainScan,
     sparse_chain, BlockId, ConfirmationTime,
 };
-use esplora_client::{BlockingClient, Builder};
 use std::collections::BTreeMap;
 
+pub use esplora_client;
+
 #[derive(Debug, Clone)]
 pub struct Client {
```
LLFourn, line 12: Remove this struct make it an extension trait and pass in parallel_requests.

bdk_esplora/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "bdk_esplora"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+
+bdk_chain = { path = "../bdk_chain", features = ["serde", "miniscript", "file_store"] }
```
LLFourn, line 10: this should have a version in it so you can release it.

bdk_esplora/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+esplora-client = { git = "https://github.com/rajarshimaitra/rust-esplora-client.git", branch = "get_recent_blocks"}
```
LLFourn, line 11: this must depend on a crates.io release. We have everything we need in the latest release (method names have probably changed).

LLFourn (approved):
I've done a cleanup and this is now approved.

This should be the last PR merged on this repo!

PR #198, opened 2023-02-23
[cli] Remove Broadcast trait

LLFourn:
In favor of passing a closure that can do the broadcast. ~`Broadcast` was forcing us to create a "client" too early in the cli workflow~ *this problem was orthogonal*. It also forces us to wrap stuff in `main.rs` so we can implement the trait.


evanlinjin (commented):
Concept ACK

danielabrozzoni (approved):
utACK 2a79a9058068988dfb7e289483c6f26bb3b1edc3

PR #203, opened 2023-02-28
Introduce `KeychainPersist`.

evanlinjin:
Most of this code is copied from the work in https://github.com/bitcoindevkit/bdk/pull/793

@LLFourn mentioned that it makes sense to put the `persist` module as part of the `bdk_chain` package.


evanlinjin:
Replaced by #204

PR #204, opened 2023-02-28
Keychain persist

evanlinjin:
Most of this code is copied from the work in https://github.com/bitcoindevkit/bdk/pull/793

@LLFourn mentioned that it makes sense to put the `persist` module as part of the `bdk_chain` package.

Note that this is based on the work done in #191 and #196


LLFourn:
Can we squash this down to one commit and base it on master?

LLFourn:
closed in favor of #206

PR #206, opened 2023-03-01
Introduce `KeychainPersist` (rebased on master)

evanlinjin:
We rebase on master because of @LLFourn's request.

This replaces #204


LLFourn (approved):
I did some renaming of types and modules.

I also added relative path depend back on to each item in the workspace.

